PR #45, opened 2019-04-11
Support for LSP Spec 3 initialized request handling

gayanper:
With this change we will not allow any request to be send to server until the Initialized request is sent. This make this library to work well with a LS which follows the Spec 3. The name of the new status is not ideal, please feel free to suggest new name if required.

### DIFF
--- a/src/main/java/org/wso2/lsp4intellij/client/languageserver/LSPServerStatusWidget.java
+++ b/src/main/java/org/wso2/lsp4intellij/client/languageserver/LSPServerStatusWidget.java
@@ -120,7 +120,7 @@ public class LSPServerStatusWidget implements StatusBarWidget {
                     JBPopupFactory.ActionSelectionAid mnemonics = JBPopupFactory.ActionSelectionAid.MNEMONICS;
                     Component component = t.getComponent();
                     List<AnAction> actions = new ArrayList<>();
-                    if (wrapper.getStatus() == ServerStatus.STARTED) {
+                    if (wrapper.getStatus() == ServerStatus.INITIALIZED) {
                         actions.add(new ShowConnectedFiles());
                     }
                     actions.add(new ShowTimeouts());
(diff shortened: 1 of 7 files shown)


rasika:
@gayanper This is great! Thanks for your contribution.  

I suppose state names need to be revamped. The state changes of `STOPPED -> STARTING -> STARTED_INIT_PENDING -> STARTED` seems bit odd to me. For the sake of readability, I would suggest is; `STOPPED -> STARTING -> STARTED -> INITIALIZED`

Therefore, shall we refactor `STARTED` into the `INITIALIZED`. Then add `STARTED` state as an intermediate state between `STARTING` and `INITIALIZED`. WDYT?

NipunaRanasinghe:
Yeah I agree with @rasika's suggestion. On the other hand, do we really need 4 states to represent this life cycle?? I personally think that treating both of the `STARTING` and `STARTED_INIT_PENDING` states as `STARTING` state would make sense too. @gayanper WDYT??

NipunaRanasinghe:
> Yeah I agree with @rasika's suggestion. On the other hand, do we really need 4 states to represent this life cycle?? I personally think that treating both of the `STARTING` and `STARTED_INIT_PENDING` states as `STARTING` state would make sense and will be less complicated. @gayanper WDYT??

rasika:
@NipunaRanasinghe IMO still we need that separation of states. That would make it more cleaner.

gayanper:
@NipunaRanasinghe i think we need a state to make sure that we only allow the initialized message to pass through. But STARTING can mean that the server is not yet started to receive requests from a consumers point of view. I know what you mean as well since how the code behaves we can assume that server is ready to receive requests even though the server is STARTING. But i do like @rasika suggestion. Shall we implement that if you don't see any drawbacks in this ?

gayanper:
@rasika I have added the changes as we agreed on

gayanper:
@NipunaRanasinghe can we get this merged ? Because some of my other changes are depended on this :(

NipunaRanasinghe:
Sure will do :)

PR #50, opened 2019-04-22
Add renaming Support

NipunaRanasinghe:
## Purpose
$subject.


src/main/java/org/wso2/lsp4intellij/contributors/rename/LSPRenameProcessor.java
```diff
@@ -0,0 +1,144 @@
...
+import org.wso2.lsp4intellij.utils.FileUtils;
+
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+public class LSPRenameProcessor extends RenamePsiElementProcessor {
```
rasika: Add a doc comment here. Check other classes as well.

NipunaRanasinghe: Added.

PR #52, opened 2019-04-22
Language client context and custom language client implementation.

gayanper:
To efficiently find Editor the language client context is introduce for
example which hold EditorEventManager for document URIs. With that you
can provide your own language client implementation.

Signed-off-by: gayanper <gayanper@gmail.com>


gayanper:
@rasika and @NipunaRanasinghe what do you think about these changes ?

NipunaRanasinghe:
@gayanper Small question, with your changes do we still need to allow DefaultRequestManager to be extended via ExtensionManager interface? I think we don't need to right?? :)

gayanper:
@NipunaRanasinghe sorry i didn't got your question. The `DefaultRequestManager` doesn't extend the `ExtentionManager` right ?

NipunaRanasinghe:
> @NipunaRanasinghe sorry i didn't got your question. The `DefaultRequestManager` doesn't extend the `ExtentionManager` right ?

I asked whether we want to keep `getExtendedRequestManagerFor()` in `LSPExtensionManager` since we have the access to the languageclient implementation now

gayanper:
@NipunaRanasinghe well what if a consumer want the request manager but has no need for extending the language client. IMHO it seems bit odd for some one to implement a totally different component to access the component they want, i mean after that the consumer might also need to store the request manager in a singleton context as well. Let me know what you think :)

NipunaRanasinghe:
@gayanper Yeah thats a valid concern, I'm just thinking of ways to improve the extend-ability of the library :) thanks!

PR #53, opened 2019-04-22
Add initial support for listening to server responses

gayanper:
When adding support for certain language servers, there is a need to receive server notifications. For example if the client needs to initially send a didChangeConfiguration then we need to wait until the server is initialize according to https://microsoft.github.io/language-server-protocol/specification#initialize. This server listener will help to receive such notifications from the server.

Right now it only support **InitializeResult** message, but the design can be extended support more notifications without breaking the backward compatibility.

### DIFF
--- a/src/main/java/org/wso2/lsp4intellij/IntellijLanguageClient.java
+++ b/src/main/java/org/wso2/lsp4intellij/IntellijLanguageClient.java
@@ -32,6 +32,7 @@ import com.intellij.openapi.vfs.VirtualFileManager;
 import org.apache.commons.lang3.tuple.ImmutablePair;
 import org.apache.commons.lang3.tuple.MutablePair;
 import org.apache.commons.lang3.tuple.Pair;
+import org.eclipse.lsp4j.DidChangeConfigurationParams;
 import org.wso2.lsp4intellij.client.languageserver.serverdefinition.LanguageServerDefinition;
 import org.wso2.lsp4intellij.client.languageserver.wrapper.LanguageServerWrapper;
 import org.wso2.lsp4intellij.editor.listeners.EditorListener;
@@ -341,5 +342,13 @@ public class IntellijLanguageClient implements ApplicationComponent {
 
     public static Map<String, Set<LanguageServerWrapper>> getProjectToLanguageWrappers() {
         return projectToLanguageWrappers;
+    }
+
+    public static void didChangeConfiguration(DidChangeConfigurationParams params,
+        Project project) {
+        final Set<LanguageServerWrapper> serverWrappers = IntellijLanguageClient
+            .getProjectToLanguageWrappers()
+            .get(FileUtils.pathToUri(project.getBasePath()));
+        serverWrappers.forEach(s -> s.getRequestManager().didChangeConfiguration(params));
     }
 }
--- a/src/main/java/org/wso2/lsp4intellij/client/languageserver/serverdefinition/ServerListener.java
+++ b/src/main/java/org/wso2/lsp4intellij/client/languageserver/serverdefinition/ServerListener.java
@@ -6,7 +6,7 @@ import org.jetbrains.annotations.NotNull;
 
 public interface ServerListener {
 
-  public static final ServerListener NO_OP = new ServerListener() {
+  public static final ServerListener DEFAULT = new ServerListener() {
   };
 
 
--- a/src/main/java/org/wso2/lsp4intellij/client/languageserver/serverdefinition/ServerListener.java
+++ b/src/main/java/org/wso2/lsp4intellij/client/languageserver/serverdefinition/ServerListener.java
@@ -1,0 +1,15 @@
+package org.wso2.lsp4intellij.client.languageserver.serverdefinition;
+
+import org.eclipse.lsp4j.InitializeResult;
+import org.eclipse.lsp4j.services.LanguageServer;
+import org.jetbrains.annotations.NotNull;
+
+public interface ServerListener {
+
+  public static final ServerListener NO_OP = new ServerListener() {
+  };
+
+
+  public default void initilize(@NotNull LanguageServer server, @NotNull InitializeResult result) {
+  }
+}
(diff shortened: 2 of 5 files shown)


gayanper:
@rasika @NipunaRanasinghe What do you think about this change for receiving server notifications ?

NipunaRanasinghe:
So the end goal is to provide a user API to expose server responses right?

src/main/java/org/wso2/lsp4intellij/client/languageserver/wrapper/MessageHandler.java
```diff
@@ -0,0 +1,43 @@
...
+import java.util.function.Function;
+import org.eclipse.lsp4j.InitializeResult;
+import org.eclipse.lsp4j.jsonrpc.MessageConsumer;
+import org.eclipse.lsp4j.jsonrpc.messages.Message;
+import org.eclipse.lsp4j.jsonrpc.messages.ResponseMessage;
+import org.eclipse.lsp4j.services.LanguageServer;
+import org.jetbrains.annotations.NotNull;
+import org.wso2.lsp4intellij.client.languageserver.serverdefinition.ServerListener;
+
+class MessageHandler implements Function<MessageConsumer, MessageConsumer> {
```
rasika: This class is package-protected. Is it intentional?

src/main/java/org/wso2/lsp4intellij/client/languageserver/serverdefinition/ServerListener.java
```diff
@@ -0,0 +1,15 @@
+package org.wso2.lsp4intellij.client.languageserver.serverdefinition;
+
+import org.eclipse.lsp4j.InitializeResult;
+import org.eclipse.lsp4j.services.LanguageServer;
+import org.jetbrains.annotations.NotNull;
+
+public interface ServerListener {
+
+  public static final ServerListener NO_OP = new ServerListener() {
```
rasika: I would love if it is called `INSTANCE`

src/main/java/org/wso2/lsp4intellij/IntellijLanguageClient.java
```diff
@@ -342,4 +343,12 @@ public static void removeWrapper(LanguageServerWrapper wrapper) {
     public static Map<String, Set<LanguageServerWrapper>> getProjectToLanguageWrappers() {
         return projectToLanguageWrappers;
     }
+
+    public static void didChangeConfiguration(DidChangeConfigurationParams params,
```
rasika: Let's avoid depending on LSP packages (`org.eclipse.lsp4j`) here. We can create `DidChangeConfigurationParams` inside the `RequestManager`. Is it correct to assume the intellij-plugin will invoke this method whenever there's config change?

NipunaRanasinghe: @rasika As I understood this is directly exposed as an user API so that users can dynamically trigger the `didChangeConfiguration()` (@gayanper correct me if I'm wrong). So to make things aligned, IMO this should be exposed via `IntelliJLanguageClient`.

Anyway since @rasika 's concern is also valid, we can implement this method in `RequestManager` and just call it here, just like what I've done with the timeouts API? WDYT?

src/main/java/org/wso2/lsp4intellij/client/languageserver/serverdefinition/ServerListener.java
gayanper: Well this is a empty or noop implementation which is used as default value. Is it correct to call it as INSTANCE ? what about DEFAULT ?

src/main/java/org/wso2/lsp4intellij/client/languageserver/wrapper/MessageHandler.java
gayanper: @rasika Yes since its scope is only for the wrapper package ? Will it have any value by making it as public ?

src/main/java/org/wso2/lsp4intellij/IntellijLanguageClient.java
rasika: @NipunaRanasinghe we still need to have the `IntellijLanguageClient.didChangeConfiguration()`. My concern was about `DidChangeConfigurationParams`. `IntellijLanguageClient` doesn't need to aware `org.eclipse.lsp4j.*` classes.  Since we already have used these classes in other places, please ignore my comment :)

gayanper: @rasika do you mean to add the DidChangeConfigurationParams inside the RequestManager class as a inner class ? IMO its good to have it as a sibling class ?

gayanper: @NipunaRanasinghe `didChangeConfiguration` was something i added additionally since its related to my requirement. But the original change here is to listen for server messages such as `ResponseMessage`. The `InitializeResult` is one kind of result that we will get from ResponseMessage.

NipunaRanasinghe: @rasika got your point :) +1 for keeping this as it is.

src/main/java/org/wso2/lsp4intellij/client/languageserver/serverdefinition/ServerListener.java
rasika: `DEFAULT` make sense to me

rasika:
@NipunaRanasinghe shall we merge this?

src/main/java/org/wso2/lsp4intellij/client/languageserver/serverdefinition/ServerListener.java
gayanper: @rasika changed to DEFAULT

src/main/java/org/wso2/lsp4intellij/IntellijLanguageClient.java
gayanper: @NipunaRanasinghe do i need to change anything more on this ?

NipunaRanasinghe: @gayanper  IMO you dont have to :)

NipunaRanasinghe:
@rasika Yeah I guess we are all good to merge :)

PR #57, opened 2019-04-23
Fix timeouts API

### LINKED ISSUE #56: New Timeout API doesn't work
The timeouts which are added into `org.wso2.lsp4intellij.requests.Timeout#setTimeouts` doesn't reload the constants in the class. For example this cause the LanguageServerWrapper to use the old INIT timeout which it was initialized at the class loading time. 

IMO we should have the timeouts in the map than in constants, and read timeout through a getter method which will return the overridden one if exist or otherwise the default one.

https://github.com/ballerina-platform/lsp4intellij/blob/fab02d60fa8a7f26540e5bf47152ec81b1f7238d/src/main/java/org/wso2/lsp4intellij/client/languageserver/wrapper/LanguageServerWrapper.java#L201

NipunaRanasinghe:
## Purpose
Fixes #56 and adds minor refactoring.

### DIFF
--- a/src/main/java/org/wso2/lsp4intellij/requests/Timeout.java
+++ b/src/main/java/org/wso2/lsp4intellij/requests/Timeout.java
@@ -39,7 +39,6 @@ public class Timeout {
     }
 
     public static void setTimeouts(Map<Timeouts, Integer> loaded) {
-        timeouts = loaded;
+        loaded.forEach((t, v) -> timeouts.replace(t, v));
     }
-
 }
--- a/src/main/java/org/wso2/lsp4intellij/requests/Timeout.java
+++ b/src/main/java/org/wso2/lsp4intellij/requests/Timeout.java
@@ -19,21 +19,6 @@ import java.util.Arrays;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
 
-import static org.wso2.lsp4intellij.requests.Timeouts.CODEACTION;
-import static org.wso2.lsp4intellij.requests.Timeouts.CODELENS;
-import static org.wso2.lsp4intellij.requests.Timeouts.COMPLETION;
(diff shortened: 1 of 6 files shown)


src/main/java/org/wso2/lsp4intellij/requests/Timeout.java
```diff
@@ -45,33 +30,9 @@
...
-
-    public static int SIGNATURE_TIMEOUT = timeouts.get(SIGNATURE);
-
-    public static int SHUTDOWN_TIMEOUT = timeouts.get(SHUTDOWN);
-
-    public static int SYMBOLS_TIMEOUT = timeouts.get(SYMBOLS);
-
-    public static int WILLSAVE_TIMEOUT = timeouts.get(WILLSAVE);
+    public static int getTimeout(Timeouts type) {
+        return timeouts.get(type);
```
gayanper: @NipunaRanasinghe seems like there is a issue still with the timeout API, for example if the consumer tries to override a timeout value by calling the plugins client `setTimeout(Timeouts, int)` API, then we will loose all the default timeout values. So shouldn't we try to put the timeout values to existing map when we call set `setTimeout(Timeouts, int)` ?

NipunaRanasinghe: @gayanper Fixed :) Could you please recheck now?

gayanper: Sure i will try in few hours time and update the issue.

gayanper: @NipunaRanasinghe its working

gayanper:
@NipunaRanasinghe shall we merge this to master ?

NipunaRanasinghe:
@gayanper  I'm waiting for @rasika to approve since PRs are not merge-able without at least one committer approval. :(

PR #63, opened 2019-04-26
Fix completion insertion and caret positioning

gayanper:
When creating lookup item we need to provide the lookup object and then
set the lookup string.
When processing text edit the replace must be perform if the range length
is not equal to newText length.

### DIFF
--- a/src/main/java/org/wso2/lsp4intellij/editor/EditorEventManager.java
+++ b/src/main/java/org/wso2/lsp4intellij/editor/EditorEventManager.java
@@ -807,9 +807,9 @@ public class EditorEventManager {
         LookupElementBuilder lookupElementBuilder;
 
         if (insertText != null && !insertText.equals("")) {
-            lookupElementBuilder = LookupElementBuilder.create(insertText);
+            lookupElementBuilder = LookupElementBuilder.create(insertText, "");
         } else if (label != null && !label.equals("")) {
-            lookupElementBuilder = LookupElementBuilder.create(label);
+            lookupElementBuilder = LookupElementBuilder.create(label, "");
         } else {
             return LookupElementBuilder.create((String) null);
         }
@@ -837,7 +837,8 @@ public class EditorEventManager {
         }
 
         return lookupElementBuilder.withPresentableText(presentableText).withTypeText(tailText, true).withIcon(icon)
-                .withAutoCompletionPolicy(AutoCompletionPolicy.SETTINGS_DEPENDENT);
+            .withLookupString(presentableText)
+            .withAutoCompletionPolicy(AutoCompletionPolicy.SETTINGS_DEPENDENT);
     }
 
     /**
@@ -964,7 +965,7 @@ public class EditorEventManager {
                     } else {
                         text = text.replace(DocumentUtils.WIN_SEPARATOR, DocumentUtils.LINUX_SEPARATOR);
                         if (end >= 0) {
-                            if (end - start <= 0) {
+                            if (end - start <= 0 || end - start != text.length()) {
                                 document.insertString(start, text);
                             } else {
                                 document.replaceString(start, end, text);
@@ -974,6 +975,7 @@ public class EditorEventManager {
                         } else if (start > 0) {
                             document.insertString(start, text);
                         }
+                        editor.getCaretModel().moveToOffset(start + text.length());
                     }
                     saveDocument();
                 });
--- a/src/main/java/org/wso2/lsp4intellij/editor/EditorEventManager.java
+++ b/src/main/java/org/wso2/lsp4intellij/editor/EditorEventManager.java
@@ -15,27 +15,6 @@
  */
 package org.wso2.lsp4intellij.editor;
 
-import static org.wso2.lsp4intellij.editor.EditorEventManagerBase.HOVER_TIME_THRES;
-import static org.wso2.lsp4intellij.editor.EditorEventManagerBase.POPUP_THRES;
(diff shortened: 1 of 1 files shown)


src/main/java/org/wso2/lsp4intellij/editor/EditorEventManager.java
```diff
@@ -965,7 +968,7 @@ public Runnable getEditsRunnable(int version, Iterable<TextEdit> edits, String n
                     } else {
                         text = text.replace(DocumentUtils.WIN_SEPARATOR, DocumentUtils.LINUX_SEPARATOR);
                         if (end >= 0) {
-                            if (end - start <= 0 || end - start != text.length()) {
```
NipunaRanasinghe: @gayanper I guess `&&` should be used instead of the `ll`. Am I missing something?

gayanper: @NipunaRanasinghe this is to handle the following situations
- The first condition handles a true insertion where the start and end is 0 or if the start is greater than end
- The second condition is to handle the scenario where lookupstring and actual lookup object is different. That is the editText contains text which might have parenthesis and brackets, but the lookup string wont have that.

But i think i need to check on this one more time according to the new contract you mentioned in the comment on the issue. I will check this tonight and see if we can get rid of this second condition.

NipunaRanasinghe: @gayanper Yeah its better if we can validate whether we need this rule, as when testing your fix with BAL LS, few false insertions were occurred due to the new rule :/

gayanper: @NipunaRanasinghe i think in my latest commit i have removed the second condition :)

NipunaRanasinghe: @gayanper My Bad :( Didn't notice it

src/main/java/org/wso2/lsp4intellij/editor/EditorEventManager.java
```diff
@@ -806,10 +807,12 @@ private LookupElement createLookupItem(CompletionItem item) {
         Icon icon = iconProvider.getCompletionIcon(kind);
         LookupElementBuilder lookupElementBuilder;
 
-        if (insertText != null && !insertText.equals("")) {
-            lookupElementBuilder = LookupElementBuilder.create(insertText);
-        } else if (label != null && !label.equals("")) {
-            lookupElementBuilder = LookupElementBuilder.create(label);
+        if (!Strings.isNullOrEmpty(label)) {
+            lookupElementBuilder = LookupElementBuilder.create(label, "");
```
NipunaRanasinghe: Do we have any specific reason to pass "" as the lookup element string?

gayanper: Well i think for label scenario it might not, but for EditText scenario it does create the LookupItem with the lookup item instead of the lookup string, we want to lookup string to be presentation string since thats what is used for matching the items for text at caret.

In STS4 we have EditText as ${xxx.yyy} where the presentation string becomes xxx.yyyy. I couldn't find any other way to get this working correctly. This is something i found in the original work of this plugin ;)

NipunaRanasinghe: Oh I see :) I'll check again with ballerina LS with your latest commit and merge.

gayanper: @NipunaRanasinghe did you test the latest commit with BAL LS ?

NipunaRanasinghe: @gayanper Unfortunately this still breaks the Ballerina LS completions. :/ I debugged and observed that,
- We need to set the text which should be inserted, as the lookup string.
- `withLookupString()` method seems to be not working somehow. (might be a bug)

So I'll add a fix and push to your fork branch, which will solve your text edit issue hopefully.So could you please check and verify if your issue is fixed with my changes? :)

gayanper:
@NipunaRanasinghe sure i will test the changes tonight and let you know the results.

NipunaRanasinghe:
@gayanper Just let me know after testing the latest changes with your plugin :)

gayanper:
Update the topic branch with master and added a fix for sts4 support since sts4 uses textEdits. I think we should not pass the textEdit to setInsertHandler when we use the textEdit to create the completion item. Please test my change with BAL LS @NipunaRanasinghe

src/main/java/org/wso2/lsp4intellij/editor/EditorEventManager.java
```diff
@@ -834,10 +832,9 @@ private LookupElement createLookupItem(CompletionItem item) {
 
         if (textEdit != null) {
             if (addTextEdits != null) {
-                addTextEdits.add(textEdit);
                 lookupElementBuilder = setInsertHandler(lookupElementBuilder, addTextEdits, command, label);
             } else {
-                lookupElementBuilder = setInsertHandler(lookupElementBuilder, Collections.singletonList(textEdit),
+                lookupElementBuilder = setInsertHandler(lookupElementBuilder, Collections.emptyList(),
```
NipunaRanasinghe: Shall we remove the else block here since we are creating a empty insert handler? :)

gayanper: Removed :+1:

Issue #66, opened 2019-04-29
Error when opening up the run configuration dialog


gayanper:
This happens when the run configuration dialog is opened.

```
java.lang.NullPointerException
	at java.util.concurrent.ConcurrentHashMap.get(ConcurrentHashMap.java:936)
	at java.util.concurrent.ConcurrentHashMap.containsKey(ConcurrentHashMap.java:964)
	at org.wso2.lsp4intellij.client.languageserver.wrapper.LanguageServerWrapper.connect(LanguageServerWrapper.java:278)
	at org.wso2.lsp4intellij.IntellijLanguageClient.lambda$editorOpened$0(IntellijLanguageClient.java:196)
	at com.intellij.openapi.application.impl.ApplicationImpl$1.run(ApplicationImpl.java:311)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at java.lang.Thread.run(Thread.java:748)
```

NipunaRanasinghe:
Closing hence fixed with #71

PR #71, opened 2019-05-02
Fix NPE when trying to open virual files which doesn't exist

gayanper:
Idea use virtual files for code snippets which are marked as
LightVirtualFile. These files are not part of the LS workspace.
These kind of file editors are used in TestNG Run Configuration for
example.

Fix for https://github.com/ballerina-platform/lsp4intellij/issues/66


NipunaRanasinghe:
@gayanper Thanks for your great work!

PR #72, opened 2019-05-03
Change access modifiers

NipunaRanasinghe:
## Purpose
When supporting custom LSP extensions, the developer needs more accessibility for the client components. This PR updates access modifiers of few such components.

### DIFF
--- a/src/main/java/org/wso2/lsp4intellij/client/languageserver/requestmanager/DefaultRequestManager.java
+++ b/src/main/java/org/wso2/lsp4intellij/client/languageserver/requestmanager/DefaultRequestManager.java
@@ -87,13 +87,14 @@ import java.util.concurrent.CompletableFuture;
 public class DefaultRequestManager implements RequestManager {
 
     private Logger LOG = Logger.getInstance(DefaultRequestManager.class);
+
+    public LanguageServerWrapper wrapper;
+    public LanguageServer server;
+    public LanguageClient client;
+    public ServerCapabilities serverCapabilities;
     private TextDocumentSyncOptions textDocumentOptions;
     private WorkspaceService workspaceService;
     private TextDocumentService textDocumentService;
-    private LanguageServerWrapper wrapper;
-    private LanguageServer server;
-    private LanguageClient client;
-    private ServerCapabilities serverCapabilities;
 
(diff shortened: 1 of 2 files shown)


nixel2007:
No getters?

src/main/java/org/wso2/lsp4intellij/client/languageserver/requestmanager/DefaultRequestManager.java
```diff
@@ -87,13 +87,14 @@
 public class DefaultRequestManager implements RequestManager {
 
     private Logger LOG = Logger.getInstance(DefaultRequestManager.class);
+
+    public LanguageServerWrapper wrapper;
+    public LanguageServer server;
+    public LanguageClient client;
+    public ServerCapabilities serverCapabilities;
```
rasika: If there's no special reason; shall we make these fields `private` and add some **getters** for these. I believe these values are being passed through the **constructor**. Allowing them `public` will make it harder to debug and test the code.

src/main/java/org/wso2/lsp4intellij/editor/EditorEventManager.java
```diff
@@ -172,10 +172,10 @@
     public Editor editor;
     public LanguageServerWrapper wrapper;
     public List<String> completionTriggers;
-    protected Project project;
-    private RequestManager requestManager;
+    public Project project;
+    public RequestManager requestManager;
+    public TextDocumentIdentifier identifier;
```
rasika: Same goes here. Keep them as **private** and add **getters**.

NipunaRanasinghe:
> Not getters?

@nixel2007 Just following YAGNI :D Will add getters anyway

src/main/java/org/wso2/lsp4intellij/client/languageserver/requestmanager/DefaultRequestManager.java
NipunaRanasinghe: Added :)

src/main/java/org/wso2/lsp4intellij/editor/EditorEventManager.java
NipunaRanasinghe: Added.

Issue #48, opened 2019-04-18
Error in LSPCompletionContributor when completing git messages in commit dialog


gayanper:
Triggering completion git dialog throws the following error
```
java.lang.NullPointerException
	at java.util.concurrent.ConcurrentHashMap.get(ConcurrentHashMap.java:936)
	at org.wso2.lsp4intellij.editor.EditorEventManagerBase.forUri(EditorEventManagerBase.java:93)
	at org.wso2.lsp4intellij.contributors.LSPCompletionContributor.invokeAutoPopup(LSPCompletionContributor.java:56)
	at com.intellij.codeInsight.editorActions.TypedHandler.isAutoPopup(TypedHandler.java:290)
	at com.intellij.codeInsight.editorActions.TypedHandler.autoPopupCompletion(TypedHandler.java:273)
	at com.intellij.codeInsight.editorActions.TypedHandler.lambda$execute$0(TypedHandler.java:180)
	at com.intellij.openapi.editor.impl.CaretModelImpl.lambda$runForEachCaret$3(CaretModelImpl.java:354)
	at com.intellij.openapi.editor.impl.CaretModelImpl.doWithCaretMerging(CaretModelImpl.java:453)
	at com.intellij.openapi.editor.impl.CaretModelImpl.runForEachCaret(CaretModelImpl.java:346)
```
Seems like the LSPCompletionContributor need some guarding.

NipunaRanasinghe:
@gayanper Thanks for reporting! :) is this intermittent or if not, could you provide a way to reproduce the issue??

gayanper:
*Steps to reproduce*
- Install the lsp client plugin, in my case it was the WIP sts4 plugin
- without triggering the lang server to startup by opening any files, open git commit dialog. 
- press ctrl+space to trigger auto completion

I will be able to provide a fix tonight 🙂

NipunaRanasinghe:
Closing since fixed by #73. Credit goes to @gayanper once more :)

PR #73, opened 2019-05-03
Fix NPE in commit dialog

gayanper:
A NPE is thrown when typing the commit message in commit dialog due to
LS client trying to resolve the dummy file uri to a EditorEventManager

Fix for https://github.com/ballerina-platform/lsp4intellij/issues/48

### DIFF
--- a/src/main/java/org/wso2/lsp4intellij/contributors/LSPCompletionContributor.java
+++ b/src/main/java/org/wso2/lsp4intellij/contributors/LSPCompletionContributor.java
@@ -52,6 +52,10 @@ class LSPCompletionContributor extends CompletionContributor {
 
     @Override
     public boolean invokeAutoPopup(@NotNull PsiElement position, char typeChar) {
+        if (!FileUtils.isFileSupported(position.getContainingFile().getVirtualFile())) {
+            return false;
+        }
+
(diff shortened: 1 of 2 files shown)


src/main/java/org/wso2/lsp4intellij/contributors/LSPCompletionContributor.java
```diff
@@ -52,6 +52,10 @@ public void fillCompletionVariants(CompletionParameters parameters, CompletionRe
 
     @Override
     public boolean invokeAutoPopup(@NotNull PsiElement position, char typeChar) {
+        if (!FileUtils.isFileSupported(position.getContainingFile().getVirtualFile())) {
+            return false;
```
NipunaRanasinghe: Shall we declare a common variable for the virtual file instance to avoid invoking `position.getContainingFile().getVirtualFile()` twice? :)

gayanper: @NipunaRanasinghe good that you pointed out, i think i have missed it after trying out the fix to do the cleanups. I think its there is no performance penalty, but from code quality point of its always good to reduce duplicates :)

gayanper: I have pushed the changes.

PR #77, opened 2019-05-10
Fix triggering LS from different windows

gayanper:
When a file is changed, the change message is sent to all LS wrappers
from all windows. With the fix now it only send to LS wrappers which
are connected to the projects the changed file belongs to.


gayanper:
@NipunaRanasinghe Have you had time to look into this PR ?

rasika:
Yes as discussed let's add this fix to imrpove the correctness. Let's create an issue(linked to original issue) to track the performance concern.

PR #78, opened 2019-05-10
Fix NPE at LSPInspection when the virtual file is null.

gayanper:
Add a file support check in addition the extension check to avoid passing down null virtual files.


gayanper:
I need your help @NipunaRanasinghe to test this since i don't have inspections in my LS.

PR #84, opened 2019-05-24
Add support for async and cancellable completion proposal contributor

gayanper:
When a LS takes lot of time for completion request the IDE freezes due to the fact the IDE cannot cancel that request according to the previous implementation. This cause the completion while typing to cause UI freeze while typing letters. With the fix the request is cancelled if the current progress is cancelled by the IDE. Also added support for providing a PsiElementPattern by the consumer so that the Contributions can be targeted to specific locations in the while by sub classing this contributor to suit consumer language.

### DIFF
--- a/src/main/java/org/wso2/lsp4intellij/contributors/LSPCompletionContributor.java
+++ b/src/main/java/org/wso2/lsp4intellij/contributors/LSPCompletionContributor.java
@@ -17,11 +17,18 @@ package org.wso2.lsp4intellij.contributors;
 
 import com.intellij.codeInsight.completion.CompletionContributor;
 import com.intellij.codeInsight.completion.CompletionParameters;
+import com.intellij.codeInsight.completion.CompletionProvider;
 import com.intellij.codeInsight.completion.CompletionResultSet;
+import com.intellij.codeInsight.completion.CompletionType;
+import com.intellij.openapi.application.ex.ApplicationUtil;
 import com.intellij.openapi.diagnostic.Logger;
 import com.intellij.openapi.editor.Editor;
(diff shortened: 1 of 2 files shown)


NipunaRanasinghe:
@gayanper As per the current implementation, if a LS takes lot of time on completion request, request should be failed after the completion timeout (which is 1000 milliseconds by default and can be changed via client API). Therefore do we really need to add support to cancel the completion requests manually? :)  @rasika WDYT?

gayanper:
When completion are enabled to show up as you type, the IDEA is cancelling the completion contributors forcefully so that the typing doesn't get blocked by this, if we keep 1 sec delay to LS completion that will effect the typing. So we need to support the cancellable completions so that IDEA can forcefully cancel the contributor. This will only happen when completion are running while you type. When you explicitly trigger it the LS timeout will govern it.

NipunaRanasinghe:
Anyway if this implementation has any performance advantages over the previous one, I think it would be good enough to proceed with this. :)

NipunaRanasinghe:
> When completion are enabled to show up as you type, the IDEA is cancelling the completion contributors forcefully so that the typing doesn't get blocked by this, if we keep 1 sec delay to LS completion that will effect the typing. So we need to support the cancellable completions so that IDEA can forcefully cancel the contributor. This will only happen when completion are running while you type. When you explicitly trigger it the LS timeout will govern it.

Oh my bad.. Thanks for the explanation! +1 for proceeding with this. I'll test this locally and merge afterwards.

NipunaRanasinghe:
@gayanper I tested it and works great! :) Thanks for working on this!

PR #85, opened 2019-05-24
Enhance support for language id passing when document open

gayanper:
With the previous implementation when supporting for multiple language id
which are not same as the ext we need to define multiple language server
definitions which cause multiple language servers to start.

With new implementation we can support for multiple languages with one
server. This also enable proper language id passing when using
extension patterns.

@NipunaRanasinghe i think this PR is really important for us since this is a breaking change if we plan this after 1.0.0

Example usage:
```
Map<String, String> langMapping = new HashMap<String,String>
langMapping.put("java", LANG_ID_JAVA)
langMapping.put("yaml", LANG_ID_YAML)
langMapping.put("yml", LANG_ID_YAML)
langMapping.put("xml", LANG_ID_XML)
langMapping.put("properties", LANG_ID_PROPERTIES)

RawCommandServerDefinition def = new RawCommandServerDefinition("java,application.*\.yaml,application.*\.yml,.*[Cc]ontext.*\.xml,application.*\.properties", langMapping, commands);

```


NipunaRanasinghe:
@gayanper Yeah you are right. We need to stabilize finalize the client API most importantly before hitting  1.0. Really appreciate your work towards this :) We need to update the documentation so I'll create an separate issue for now and will add the fix with the next documentation PR.

PR #86, opened 2019-05-26
Enhance support for LSP inspections and update documentation

gayanper:
Now the LSP inspection is not manually triggered, this enhance the
performance of the LSP plugin since the inspections are triggered once
the IDE ready to execute them. This also avoid hacking the way through
to execute inspections by force.

Signed-off-by: gayanper <gayanper@gmail.com>

@NipunaRanasinghe please test this with BAL LS. You will need to enable the LSP inspection from the inspection settings.


NipunaRanasinghe:
@gayanper This is the way how inspections works in our good old Scala plugin ;) I've tested this way and found a critical design issue which is shown below.

Currently we keeps the list of diagnostics in the `EditorEvenetManager` instance and using that list in `LSPInspection`. Therefore If the language server diagnostic notification does not receive before the `checkFile()` invocation in `LSPInspection` (which is automatically triggered by IDEA platform), previous set of diagnostics will be shown instead of the correct diagnostics.So that is why I had to trigger inspections manually, only after getting the diagnostics notification from the language server. 

Do you have any ideas to overcome the issue I mentioned apart from triggering the inspections manually? :)

gayanper:
@NipunaRanasinghe i will see if we can ask the idea to trigger a inspection or debounce if a inspection is already requested.

By the way there are some more design issues in gitter can you check those as well ?

gayanper:
@NipunaRanasinghe try the new fix i made to support delayed diagnostics

NipunaRanasinghe:
@gayanper I tried with you latest commit but doesn't work fine and noticed a little performance degradation as well :/ Please refer the following example to see how the inspections are not applied correctly and being corrected in the next file change event as I explained in the first place.
inspections-issue

gayanper:
I see the problem, but there shouldn't be a performance degration. because forcefully executing inspections with language which has lot of inspection every time a diagnosis change is received is far more worse than we holding of the inspection thread for few seconds if we have no diagnosis received. May be the problem might be in the way the completable list object is handling this.

May be as a different approach we need to find a way to run only the LSP inspection on each diagnosis message. That might solve the problem and not put a heavy burden on the IDE by executing all the inspections and external inspection tools like Sonar on each LSP diagnosis message.

gayanper:
May be you can try using a Lock in CompletableListValue methods, because there is a chance that already completed CF can be updated because it is assigned with a new one. If we acquire a lock after we read the value from CF until we assigned the new CF we can avoid it the reassignment. To try this theory can you just make the methods synchronize and see if there is any change in the behavior @NipunaRanasinghe  ?

gayanper:
Try the new implementation to see if it solves the issue, in the long run we should try the Annotator or ExternalAnnotator after some experimentation. The new implementation use rxjava to create a stream of events which debounce, if this works i will try to use the observable to actually carry the diagnostics list. Then we can also avoid unnecessary mutations of the diagnostics list as well.

NipunaRanasinghe:
@gayanper I've already tried using debouncing before moving to the forced inspection triggering and the reason I didn't use it was the noticeable delay it caused( total delay caused is diagnostics notification delay + debounce delay). And since you are not using the forced inspection invalid diagnostics set is applied to the document as soon after a document change and until the debounce implementation set the correct inspections (say after 1s) the invalid inspection set is visible to the user. 

Anyway as you said I was also thinking of replacing the current implementation by writing an annotator for LSP diagnostics :) Hopefully I'll be able to have a look on it after my deadlines.

gayanper:
Seems like the bal ls is some what slow on diagnostics than the sts4 ls. Because i don't see any issue in sts4. By the way do we really need to be that much synched ? I mean event in Idea it takes a while to remove and reaply new inspections 😉

gayanper:
Basically what your forced inspection doing is slowing down the user from typing until the inspections are applied. But idea inspections nor the annotator is not meant to work like that.

NipunaRanasinghe:
> Seems like the bal ls is some what slow on diagnostics than the sts4 ls. Because i don't see any issue in sts4. By the way do we really need to be that much synched ? I mean event in Idea it takes a while to remove and reaply new inspections

Yeah it does seem that the ballerina LS diagnostics generation is comparatively slow than the other languages and hopefully it will get better after the ongoing compiler re-writing process :) Anyway IMO we should be aware of that scenario and go for a better option if needed (I hope that the `ExternalAnnotator` might be the solution)

gayanper:
@NipunaRanasinghe i think i will try the annotator approach and we will see. I also hope it might solve the issue. Fingers cross though.

gayanper:
@NipunaRanasinghe try the new changes with External Annotator support. To try to ballerina you will need to introduce a langauge in your plugin since Annotators need a language to work. I tried using "any" and "TEXT" but it didn't workout as a base Language.

matthewmichihara:
Looks like some of the problems I was having is fixed by this PR, so thanks!

Out of curiosity, what's going on here?
```
        if (extToServerDefinition.keySet().contains(ext)) {
            return true;
        }
        // try to see if the extension matches a extension pattern in the definition map.
        return extToServerDefinition.keySet().stream().filter(p -> p.endsWith(ext)).findFirst().isPresent();
```

gayanper:
@matthewmichihara That is for handling the file name patterns. Since we still don't have the language support so that we can define the langeservice definition for a intellij language we have to use file pattern some times.
For example in STS4 plugin, i use `.*context.xml` file pattern for matching spring xml configurations. otherwise the STS4 language server will be bound to all xml files.

gayanper:
Fixed compilation issue at isExtensionSupported, now every thing should be working

gayanper:
@NipunaRanasinghe seem like we have master failure

NipunaRanasinghe:
@gayanper Yeah master was failing due to a missing license header and I fixed it. Could you please re-sync your branch with the master branch?

gayanper:
@NipunaRanasinghe done

NipunaRanasinghe:
@gayanper I think we are on the right path now and it seems to be outperforming our previous hacky implementation ;)  

Unfortunately I can still see the invalid diagnostics for ballerina LS due to the issue I previously mentioned. So we can either use a debouncer or we have to find a way to trigger the annotator manually only after getting the diagnostics from the LS, instead of triggering it on each document content change event. IMO second option might be the optimum solution since it be better than hard-coding a debounce interval.

Apart from that, we have to bind the code action support here, like we have done in the previous implementation.

gayanper:
I think i don't have a LS server to test this and i'm little bit busy these days. if you need i can provide you access to work on my branch or you can merge this and work on the code action after that. WDYT ?

nixel2007:
Guys, i didn't read all thread, but if you need some testing, I can test branch with BSL LS. It has diagnostics and code actions (as quick fixes).
So if you provide step-by-step test scenario and what you expect to see with LS, I can do it for you.

gayanper:
I found a little time which made me to implements the code actions. Please try it out @NipunaRanasinghe and @nixel2007

NipunaRanasinghe:
@gayanper I tested your PR and works fine apart from the issue I have been mentioning from the beginning :) I'll try to add debouncing or some other appropriate fix to get this "done done".

dinvlad:
Hi folks, any updates on when this will be merged? Alas, we can't see any diagnostics, and this appears to address that (however, I wasn't able to see any diagnostics even using `com.github.gayanper:lsp4intellij:pr~enhance-inspections-SNAPSHOT` build).

gayanper:
@dinvlad did you add the new annotator against the language ? You need to define a language for your source file if intellij doesn't has one for your files. Check this for more details https://intellij-support.jetbrains.com/hc/en-us/community/posts/360001560400-Annotator-vs-ExternalAnnotator

You need to add the LSPAnnotator as a External Annotator in your plugin.xml for your language.

dinvlad:
@gayanper yep, added the external annotator and a language definition as described in http://www.jetbrains.org/intellij/sdk/docs/tutorials/custom_language_support/language_and_filetype.html, but that didn’t help. I feel like we’re missing something else that just completely prevents any LSP diagnostics from showing up.

dinvlad:
Here's our config, for the reference: https://github.com/broadinstitute/wdl-ide/blob/test-annotations/client/intellij/src/main/resources/META-INF/plugin.xml

dinvlad:
So I've done a bit more investigation on this - it looks like the annotator is partially working, but it either:

1) doesn't always report the diagnostics (e.g. I have to add a space to my code so it then re-evaluates the code and only `then` publish a diagnostic, or

2) it crashes the plugin with the following stack trace:
```
com.intellij.diagnostic.PluginException: annotator: org.wso2.lsp4intellij.contributors.annotator.LSPAnnotator@501306a (class org.wso2.lsp4intellij.contributors.annotator.LSPAnnotator) [Plugin: org.broadinstitute.wdl.devtools]
	at com.intellij.ide.plugins.PluginManagerCore.createPluginException(PluginManagerCore.java:368)
	at com.intellij.diagnostic.PluginProblemReporterImpl.createPluginExceptionByClass(PluginProblemReporterImpl.java:12)
...
Caused by: java.util.ConcurrentModificationException
	at java.util.ArrayList.forEach(ArrayList.java:1260)
	at org.wso2.lsp4intellij.contributors.annotator.LSPAnnotator.createAnnotations(LSPAnnotator.java:76)
	at org.wso2.lsp4intellij.contributors.annotator.LSPAnnotator.apply(LSPAnnotator.java:66)
	at com.intellij.codeInsight.daemon.impl.ExternalToolPass.doApply(ExternalToolPass.java:206)
	... 37 more
```

So it looks like there might be some concurrency issues that prevent it from working correctly..

dinvlad:
When it does show diagnostics, they "lag behind" for large files. I can introduce a syntax error, which is not reported right away (no matter how long I wait). Then I fix the error with "Undo", and this time the error is displayed, even though it was just fixed. This seems to not happen for smaller files. I guess this is the issue @gayanper referred to earlier?

dinvlad:
In addition, I’d like to chime in on the necessity to define a language. We plan to use TextMate Bundles plugin to enable syntax highlighting, and that collides with custom language registration required for annotations, because TextMate plugin has its own registration system that refuses to add an already added language. How should we resolve this conflict?

NipunaRanasinghe:
> When it does show diagnostics, they "lag behind" for large files. I can introduce a syntax error, which is not reported right away (no matter how long I wait). Then I fix the error with "Undo", and this time the error is displayed, even though it was just fixed (the error goes away only after I edit the file once more). This seems to not happen for smaller files. I guess this is the issue @NipunaRanasinghe referred to earlier?

@dinvlad Yeah it happens since LSP diagnostics notifications and Annotator are not synchronized. I'll find a way if it possible to trigger the annotator only after receiving each LS diagnostics notification. Hopefully will be able to have a look after our ongoing language release work :)

NipunaRanasinghe:
@dinvlad could you please check whether your aforementioned issue is resolved with our latest fixes? :)

dinvlad:
@NipunaRanasinghe thanks, is there an easy way to try it? In the past, I've declared
```
dependencies {
    implementation 'com.github.gayanper:lsp4intellij:pr~enhance-inspections-SNAPSHOT'
    ...
}
```
but that seems no longer working (i.e. it's unable to resolve the dependency on Jetpack).

dinvlad:
Update: just resolved it with
```
implementation 'com.github.gayanper:lsp4intellij:e12a9280264ff92f3e753774d8ae30c7f749602e'
```
and the annotations are now displayed correctly!!

I'll be testing it a bit more to make sure everything works as expected.
Thanks!!

NipunaRanasinghe:
@dinvlad Perfect! Thanks for testing and please let us know if anything goes wrong :) We'll do a new stable release after merging.

dinvlad:
Hmm, I don't see `org.wso2.lsp4intellij.contributors.annotator.LSPAnnotator` class in `lsp4intellij-0.92.1.jar`. Is it possible that the build didn't include it?

NipunaRanasinghe:
@dinvlad Yeah actually 0.92.1 is an older release which I added the release notes recently. We can do another release after merging the remaning PR :)

dinvlad:
Alright, so I played with it a bit more and noticed one issue I've seen earlier:
one has to disable Editor -> Inspections -> Spelling -> Typo inspection, otherwise it interferes with lsp4intellij inspections and the latter are not shown.

Could this be solved programmatically with a setting for the LSP plugin?

dinvlad:
Found a workaround:
1) Add to plugin.xml:
```xml
        <postStartupActivity
                implementation="org.example.plugin.PluginStartupActivity"
        />
```
2) Implement
```java
public final class PluginStartupActivity implements StartupActivity {
    public void runActivity(@NotNull Project project) {
        InspectionProjectProfileManager.getInstance(project)
                .getCurrentProfile().disableToolByDefault(
                    Collections.singletonList("SpellCheckingInspection"), project
        );
    }
}
```

gayanper:
@dinvlad what is your lsp plugin is about ? What does it do ? Because with my sts4 plugin I don't have a issue with spell checker. Theoritically there shouldn't be any conflict.

dinvlad:
@gayanper - it’s a plugin for a DSL called WDL (Workflow Description Language)[1].

With Typo enabled, I get the same behavior as described at the very beginning of this issue - the LSP diagnostics appear briefly, but then disappear as they seem to be “replaced” by the diagnostics from Typos.

[1] https://github.com/broadinstitute/wdl-ide/tree/master/client/intellij

PR #104, opened 2019-06-26
Add signature help support

NipunaRanasinghe:
## Purpose
$subject.

### DIFF
--- a/src/main/java/org/wso2/lsp4intellij/editor/EditorEventManager.java
+++ b/src/main/java/org/wso2/lsp4intellij/editor/EditorEventManager.java
@@ -294,50 +294,53 @@ public class EditorEventManager {
      * @param e the event
      */
     public void mouseMoved(EditorMouseEvent e) {
-        if (e.getEditor() == editor) {
-            Language language = PsiDocumentManager.getInstance(project).getPsiFile(editor.getDocument()).getLanguage();
-            if ((LanguageDocumentation.INSTANCE.allForLanguage(language).isEmpty() || language
-                    .equals(PlainTextLanguage.INSTANCE)) && (getIsCtrlDown() || EditorSettingsExternalizable
-                    .getInstance().isShowQuickDocOnMouseOverElement())) {
-                long curTime = System.nanoTime();
-                if (predTime == (-1L) || ctrlTime == (-1L)) {
-                    predTime = curTime;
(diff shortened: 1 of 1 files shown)


src/main/java/org/wso2/lsp4intellij/editor/EditorEventManager.java
```diff
@@ -183,6 +186,8 @@
     private EditorMouseListener mouseListener;
     private EditorMouseMotionListener mouseMotionListener;
 
+    public List<String> completionTriggers;
+    public List<String> signatureTriggers;
```
nixel2007: should it be List<Character> ?

NipunaRanasinghe: lsp4j has defined both completion-triggers and signature-triggers as `List<String>` (Not sure the reason why) so I preserved the type as it is.

src/main/java/org/wso2/lsp4intellij/editor/EditorEventManager.java
```diff
@@ -259,7 +268,9 @@ public DidChangeTextDocumentParams getChangesParams() {
      * @param c The character just typed
      */
     public void characterTyped(char c) {
-        // Todo - Implement
+        if (signatureTriggers.contains(Character.toString(c))) {
```
nixel2007: As i can see, this code expects one-symbol string, cause `c` is a char.
@NipunaRanasinghe

nixel2007: so... it could be converted to List<Character> i guess

NipunaRanasinghe: Yeah that is one option :) We can either convert the List<String> to List<Character> when reading the server options or we can keep it as it is and do the check like this.. Just a matter of where the conversion should be done right? Which one do you prefer?

nixel2007: @NipunaRanasinghe please, add ` around your Lists :D

NipunaRanasinghe: @nixel2007 oh my bad :smile: done!

PR #114, opened 2019-07-21
fix a file reformat bug. 

flyindirt:
fix bug when reformat a cpp file or code selection when lsp server is clangd.
add parameters in FormattingOptions  construction  to avoid equest resolve exception in clangd, as in scala version intellij-lsp project.


NipunaRanasinghe:
@flyindirt First of all, forgive me for not being around here for a while and thank you so much for your contribution! :) However I think its better not to have hard-coded configurations in the client side since the default constructor seems to be allowed by the LSP.  So IMO the server side should be fail proof and apply the language specified formatting options in case the client request does not specify it. 

And if you want to send custom formatting options as per your requirement, you can simply override the required method in by extending `EditorEventManager` and provide your extended editor event manager using `LSPExtensionManager` interface. (we are still working on adding documentation for  LSP extension capabilities, so until that you can refer our ballerina plugin as an example)

PR #115, opened 2019-07-31
tiny change to Reformat , so that it will be the same user experience as Java in Intellij IDEA

sonicning:
1.  add reformat selection to LSPReformatAction
2.  do not pop out show reformat dialog twice

## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests
   > Details about the test cases and coverage

## Security checks
 - Followed secure coding standards in http://wso2.com/technical-reports/wso2-secure-engineering-guidelines? yes/no
 - Ran FindSecurityBugs plugin and verified report? yes/no
 - Confirmed that this PR doesn't commit any keys, passwords, tokens, usernames, or other secrets? yes/no

## Test environment

openjdk version "11.0.3" 2019-04-16
OpenJDK Runtime Environment (build 11.0.3+12-b304.10)
OpenJDK 64-Bit Server VM (build 11.0.3+12-b304.10, mixed mode)

$ ./clangd --version
clangd version 8.0.0 (tags/RELEASE_800/final)

### DIFF
--- a/src/main/java/org/wso2/lsp4intellij/actions/LSPReformatAction.java
+++ b/src/main/java/org/wso2/lsp4intellij/actions/LSPReformatAction.java
@@ -45,7 +45,12 @@ public class LSPReformatAction extends ReformatCodeAction implements DumbAware {
         PsiFile file = PsiDocumentManager.getInstance(project).getPsiFile(editor.getDocument());
         if (LanguageFormatting.INSTANCE.allForLanguage(file.getLanguage()).isEmpty() && IntellijLanguageClient
                 .isExtensionSupported(file.getVirtualFile())) {
-            ReformatHandler.reformatFile(editor);
+            // if editor hasSelection, only reformat selection, not reformat the whole file
+            if (editor.getSelectionModel().hasSelection()) {
+                ReformatHandler.reformatSelection(editor);
+            } else {
+                ReformatHandler.reformatFile(editor);
+            }
         } else {
             super.actionPerformed(e);
         }
--- a/src/main/java/org/wso2/lsp4intellij/actions/LSPShowReformatDialogAction.java
+++ b/src/main/java/org/wso2/lsp4intellij/actions/LSPShowReformatDialogAction.java
@@ -66,7 +66,8 @@ public class LSPShowReformatDialogAction extends ShowReformatFileDialog implemen
                         }
                     }
                 } else {
-                    super.actionPerformed(e);
+                    // if user chose cancel , the dialog in super.actionPerformed(e) will show again
+                    // super.actionPerformed(e);
                 }
             } else {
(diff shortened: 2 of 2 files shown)


NipunaRanasinghe:
@sonicning Thank you so much for working on this! Your PR looks all good and the there are two things that I'm worrying about

1. IMO if a particular language server does not support partial(range) formatting (we can check this using the received server capabilities), the client should falls back to full file formatting to preserve the user experience.
2. There might be cases where there are "unintentional" selections in the file that the user is not aware of.

So I would like to suggest that its better to show the reformat dialog if any there are any selections in the file so that the user can choose their preferred action. WDYT?

sonicning:
@NipunaRanasinghe 
1.  Yes, I haven't thought about that for point 1. I am using clangd as language server, so it is functional. What kind of error will it be when the language server does not support partial formatting ? Will this line of code throw execption ? or just do nothing ? 

2. I think we can treat "unintentional" selections as "intentional" selections. As in Intellij IDEA, that is the case for Java. If there is selection, no matter "unintentional" or "intentional", IDEA will reformat selection code.

rasika:
> 1.  IMO if a particular language server does not support partial(range) formatting (we can check this using the received server capabilities), the client should falls back to full file formatting to preserve the user experience.
If we look at Java experience in IntelliJ, it does nothing when it fails to format. I suppose we can do the same. Here, if the server does not support partial formatting if will be communicated at the initial stage of the LSP protocol via server capabilities. Even after that, if the client sends a partial formatting request we can simply ignore it. As a user, falling back into the full file formatting is not the accepted behavior for me. Rather I would expect do nothing when if fails to format. 

> 2.  There might be cases where there are "unintentional" selections in the file that the user is not aware of.
Agree with @sonicning IMO We don't need to add extra steps into formatting. This should be the same as Java partial formatting.

NipunaRanasinghe:
>Yes, I haven't thought about that for point 1. I am using clangd as language server, so it is functional. What kind of error will it be when the language server does not support partial formatting ? Will this line of code throw execption ? or just do nothing ?

If you could have a look at the `DefaultRequestManager` you can see we are checking whether the connected language server has the particular capability before sending the request :) So clangd won't receive the range formatting request if it doesn't support the partial formatting.

> I think we can treat "unintentional" selections as "intentional" selections. As in Intellij IDEA, that is the case for Java. If there is selection, no matter "unintentional" or "intentional", IDEA will reformat selection code.

Yeah that is the default behavior for java in IDEA. I was saying is we can add an improvement where if the attached language server does not support partial formatting we can simply notify it to user and ask to format the file only if the user wants. But I think we can think about it after merging this since its not a blocker for now :)

sonicning:
@NipunaRanasinghe Yes, Good idea, I will look into that when I have time, thanks

Issue #121, opened 2019-08-16
Goto Definition does not "go" to definition


dinvlad:
**Description:**
Command-clicking on a symbol should "_jump_ to definition", if implemented by LSP.

**Affected Product Versions:**
IntelliJ 2019.1.3, 2019.2.1 (EAP)

`com.github.gayanper:lsp4intellij:e12a9280264ff92f3e753774d8ae30c7f749602e`

**OS, DB, other environment details and versions:**    
Mac OS 10.14.6

**Steps to reproduce:**
1. Implement `textDocument/definition` handler in the language server.
2. Verify that when I cmd+click on a symbol in VS Code, it "jumps" to the definition within the same file - i.e., it moves the viewport to the correct line of code.
3. Attempt to use this feature in IntelliJ.
4. Instead of "jumping" to the definition within the _same_ file, IJ just _selects_ it, without moving the viewport. If the target file is different, then it does "jump" to that file (and still _selects_ the definition).

As a user, I'd like it to:
1. Change the viewport when "jumping" within the same file.
2. Prefer NOT to select the target definition, because that provides an eyesore when the target is a "block" of code instead of a particular word. Instead, I'd like the cursor to physically jump to the beginning of the target word/block.

**Related Issues:**
I'm using @gayanper's fork from https://github.com/ballerina-platform/lsp4intellij/pull/86, if that matters.

I'd be happy to work on a PR, if you could suggest where the culprit might be.
Thanks!

dinvlad:
From a brief check, it seems that the offending line
https://github.com/ballerina-platform/lsp4intellij/blob/884c517bf63f0731905b3a14b3ce678a329b7f42/src/main/java/org/wso2/lsp4intellij/editor/EditorEventManager.java#L377

should indicate the position to jump to, instead of just the file. I'll create a PR.

PR #122, opened 2019-08-16
Navigate to line/column in goto definition/references

dinvlad:
Per https://github.com/ballerina-platform/lsp4intellij/issues/121, command-clicking on a symbol should "_jump_ to definition", if implemented by LSP.

Instead, currently it only selects the target range in the editor, without navigating to it.

Navigation to the target range on ctrl+click is common in other plugins for IntelliJ, as well as other editors like VS Code. We're just enabling it here.

Additionally, we remove "selection" for the target range, which provides an eyesore if the selection is large (e.g. when the target is a block of code).

This PR adds line/column arguments to `OpenFileDescriptor()` constructor, which is used by `openTextEditor()` call to open the target range.

The target line/column is taken to be the first line/char of the target range.

Additionally, it removes previous caret/selection manipulation code that came right after,
but didn't serve the intended purpose (in my view).

out

Navigate to line/column in goto definition/references

### DIFF
--- a/src/main/java/org/wso2/lsp4intellij/editor/EditorEventManager.java
+++ b/src/main/java/org/wso2/lsp4intellij/editor/EditorEventManager.java
@@ -375,17 +375,9 @@ public class EditorEventManager {
                 }
                 if (file != null) {
                     final Position start = loc.getRange().getStart();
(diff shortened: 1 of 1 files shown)


dinvlad:
Awesome, thanks @NipunaRanasinghe!

PR #133, opened 2019-09-13
Update symbol contributor to new API

gayanper:
This fixed handling of duplicate names in symbol list if two items
have the same name.

Before when we have duplicate names the items shown is 2 * number duplicates. This fixes it.


gayanper:
@NipunaRanasinghe the previous API runs through a adapter which has a default way of mapping navigationitem and names. Intellij core ia actually working with this new extension. Hope i answer your question.

NipunaRanasinghe:
> @NipunaRanasinghe the previous API runs through a adapter which has a default way of mapping navigationitem and names. Intellij core ia actually working with this new extension. Hope i answer your question.

Thanks for the explanation! :)

PR #134, opened 2019-09-14
Fix not sending didChangeWatchedFiles for open document.

gayanper:
Update the file notifications according to the discussion done here
https://github.com/Microsoft/language-server-protocol/issues/60. Some
of the language servers depends on this. For example STS4


gayanper:
@NipunaRanasinghe any update on this PR ?

PR #135, opened 2019-09-17
Add support for properly shutting down language servers

gayanper:
Add support for shutting down language server when IDE window is closed and IDEA jvm is forcefully stopped.


gayanper:
wrong branch

Issue #137, opened 2019-09-18
Add line-based code actions support


NipunaRanasinghe:
**Description:**
Currently we only support diagnostics bound code actions and the aforementioned feature is available in the vs code language client. Basically we will have to add a caret listener and send a code actions request each time when the caret position is changed.

**Suggested Labels:**

**Suggested Assignees:**

**Affected Product Version:**

**OS, DB, other environment details and versions:**    

**Steps to reproduce:**

**Related Issues:**

NipunaRanasinghe:
Closing since this is fixed with #137.

PR #139, opened 2019-09-20
Revamp code action implementation

NipunaRanasinghe:
## Purpose
This PR revamps the code actions implementation, which also resolves a critical performance issue related to diagnostics. Now instead of fetching code actions on each document change, code actions requests are sent only when user places the cursor on the diagnostic or on a line. Therefore this also adds #137.

### DIFF
--- a/src/main/java/org/wso2/lsp4intellij/contributors/annotator/LSPAnnotator.java
+++ b/src/main/java/org/wso2/lsp4intellij/contributors/annotator/LSPAnnotator.java
@@ -18,30 +18,27 @@ package org.wso2.lsp4intellij.contributors.annotator;
 import com.intellij.lang.annotation.Annotation;
 import com.intellij.lang.annotation.AnnotationHolder;
 import com.intellij.lang.annotation.ExternalAnnotator;
+import com.intellij.openapi.diagnostic.Logger;
 import com.intellij.openapi.editor.Editor;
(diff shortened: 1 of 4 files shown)


src/main/java/org/wso2/lsp4intellij/contributors/annotator/LSPAnnotator.java
```diff
@@ -81,23 +78,51 @@ public void apply(@NotNull PsiFile file, Object annotationResult, @NotNull Annot
...
+                } catch (ConcurrentModificationException e) {
+                    // Todo
+                } catch (Throwable t) {
+                    LOG.warn("Error occurred when updating LSP diagnostics.", t);
+                }
+            } else if (eventManager.isDiagnosticSyncRequired()) {
+                try {
+                    createAnnotations(holder, eventManager);
+                } catch (ConcurrentModificationException e) {
+                    // Todo
```
rasika: Let's add a TODO into line 99, to solve concurrent modification issue. Having it here, fails the `createAnnotations()` invocation silently whenever a `ConcurrentModificationException` occurs.

NipunaRanasinghe: +1 and  modified accordingly :)

PR #141, opened 2019-09-22
Improve symbol navigation

gayanper:
Change how the URL to file is processed with the help of IntelliJ IDEA
util classes so that the URL is IntelliJ IDEA friendly. This is really
needed for opening files which are inside archive files such as jar
files.

The previous implementation didn't support opening files inside jar files such as opening source code from source jar.


NipunaRanasinghe:
@gayanper Oh I see! I'll keep in mind to use this util in the future in uri sensitive scenarios :) Thanks for pointing this out

PR #144, opened 2019-09-26
Add support for customize icons from LS client implementation.

gayanper:
Add support for customize icons from LS client implementation (enhanced)

- Fix all the places where default static implementation was used for Icon
provider.
- Moved the IconProvider to ExtensionManager instead of the
ServerDefinition

### DIFF
--- a/src/main/java/org/wso2/lsp4intellij/extensions/LSPExtensionManager.java
+++ b/src/main/java/org/wso2/lsp4intellij/extensions/LSPExtensionManager.java
@@ -18,14 +18,18 @@ package org.wso2.lsp4intellij.extensions;
 import com.intellij.openapi.editor.Editor;
 import com.intellij.openapi.editor.event.DocumentListener;
 import com.intellij.openapi.editor.event.EditorMouseListener;
+import com.intellij.psi.PsiFile;
 import org.eclipse.lsp4j.ServerCapabilities;
 import org.eclipse.lsp4j.services.LanguageClient;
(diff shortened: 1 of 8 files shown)


src/main/java/org/wso2/lsp4intellij/extensions/LSPExtensionManager.java
```diff
@@ -49,4 +53,28 @@
...
+     * Some language servers might only need to start for files which has a specific content. This method can be used
+     * in such situation to control whether the file must be connected to a language server which is registered for the
+     * extension of this file.
+     *
+     * <b>Note:</b> By default this method returns <code>true</code>
+     *
+     * @param file PsiFile which is about to connect to a language server.
+     * @return <code>true</code> if the file is supported.
+     */
+    default boolean isFileContentSupported(@NotNull PsiFile file) {
```
NipunaRanasinghe: I'm seeing that this util method is still not used. Could you please explain why do we need this here? :) (i.e. instead of somewhere like `FileUtils`?)

gayanper: I think this change is not part of this. I will fix it tonight and update the pr

NipunaRanasinghe: Sure :) thanks

gayanper: Done please merge this and let me know, i will then merge the label provider PR

PR #149, opened 2019-10-13
Add support for quick doc hovers (support for opening files)

gayanper:
- enabled HTML formatting for all content since LSP spec is now sending
markup content
- use a textpane with JB html kit to style the documentation
- add support for lanuguage like java which doesn't inherit from plain
text language.
- add support for open source files from links in quick doc hover

quick-docs

### DIFF
--- a/src/main/java/org/wso2/lsp4intellij/editor/EditorEventManager.java
+++ b/src/main/java/org/wso2/lsp4intellij/editor/EditorEventManager.java
@@ -15,6 +15,9 @@
  */
 package org.wso2.lsp4intellij.editor;
 
+import com.google.common.cache.CacheBuilder;
+import com.google.common.cache.CacheLoader;
+import com.google.common.cache.LoadingCache;
 import com.intellij.codeInsight.completion.InsertionContext;
 import com.intellij.codeInsight.daemon.DaemonCodeAnalyzer;
 import com.intellij.codeInsight.hint.HintManager;
@@ -82,10 +85,7 @@ import java.awt.Cursor;
 import java.awt.Point;
 import java.net.MalformedURLException;
 import java.net.URL;
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.List;
-import java.util.Objects;
+import java.util.*;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
@@ -255,9 +255,8 @@ public class EditorEventManager {
             return;
         }
         Language language = psiFile.getLanguage();
-        if ((!LanguageDocumentation.INSTANCE.allForLanguage(language).isEmpty() && !language
-                .equals(PlainTextLanguage.INSTANCE)) || (!getIsCtrlDown() && !EditorSettingsExternalizable
-                .getInstance().isShowQuickDocOnMouseOverElement())) {
+        if ((!LanguageDocumentation.INSTANCE.allForLanguage(language).isEmpty() && !isSupportedLanguageFile(psiFile))
+                || (!getIsCtrlDown() && !EditorSettingsExternalizable.getInstance().isShowQuickDocOnMouseOverElement())) {
             return;
         }
 
@@ -295,6 +294,11 @@ public class EditorEventManager {
             }
             predTime = curTime;
         }
+    }
+
+    private boolean isSupportedLanguageFile(PsiFile file) {
+        return file.getLanguage().isKindOf(PlainTextLanguage.INSTANCE)
+            || FileUtils.isFileSupported(file.getVirtualFile());
     }
 
     /**
--- a/src/main/java/org/wso2/lsp4intellij/utils/GUIUtils.java
+++ b/src/main/java/org/wso2/lsp4intellij/utils/GUIUtils.java
@@ -17,9 +17,20 @@ package org.wso2.lsp4intellij.utils;
 
 import com.intellij.codeInsight.hint.HintManager;
 import com.intellij.codeInsight.hint.HintManagerImpl;
+import com.intellij.ide.browsers.BrowserLauncher;
+import com.intellij.openapi.diagnostic.Logger;
 import com.intellij.openapi.editor.Editor;
+import com.intellij.openapi.fileEditor.FileEditorManager;
+import com.intellij.openapi.fileEditor.OpenFileDescriptor;
+import com.intellij.openapi.project.Project;
+import com.intellij.openapi.ui.Messages;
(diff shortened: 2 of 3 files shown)


src/main/java/org/wso2/lsp4intellij/utils/GUIUtils.java
```diff
@@ -49,7 +67,33 @@ public static Hint createAndShowEditorHint(Editor editor, String string, Point p
...
+                        } else {
+                            final Project project = editor.getProject();
+                            VirtualFile file = VfsUtil.findFileByURL(new URL(VfsUtilCore.fixURLforIDEA(e.getURL().toString())));
+                            final OpenFileDescriptor descriptor = new OpenFileDescriptor(project, file);
+                            writeAction(() -> {
+                                FileEditorManager.getInstance(project).openTextEditor(descriptor, true);
+                            });
+                        }
+                    } catch (MalformedURLException | URISyntaxException ex) {
+                        LOGGER.error("Error opening doc link", ex);
```
NipunaRanasinghe: Shall we prompt a message to the user instead of showing logging this as an error (which will be shown as an fatal error in the IDE)? WDYT?

gayanper: Sure may be we can log this as a debug message and show a notification which will be shown in the event log. WDYT ?

NipunaRanasinghe: I'm afraid that the event log will be noisy if we continue prompting these kind of exceptions since it's shared by all the tools for prompting critical events :/  How about the IntelliJ Messages API? 

```
Messages.showErrorDialog(message, title);
```

gayanper:
I meant to just show a info event in the event log. I don't see that we will get lot of such errors 🤔. Of cause we can use the Message API but then user will need to close the popup right ? Will that be better from UX point of view ?.

src/main/java/org/wso2/lsp4intellij/utils/GUIUtils.java
gayanper: I meant to just show a info event in the event log. I don't see that we will get lot of such errors ?
Of cause we can use the Message API but then user will need to close the popup right ? Will that be better from UX point of view ?.

rasika: @gayanper @NipunaRanasinghe my suggestion is also to move ahead with the Messages API. If we are going to add this into event log; we need to treat & add all error messages in other places in same level and I am afraid that will flood the events log. IMO since this action is user-initiated (clicking the hyperlink); showing error message in a failure won't harm the UX. WDYT?

nixel2007:
Sorry for offtopic... Could you post some screenshots or gifs with $subject?

NipunaRanasinghe:
> Sorry for offtopic... Could you post some screenshots or gifs with $subject?

Yeah that would be nice

gayanper:
Do you want a gif on the quick doc with clickable links ?

NipunaRanasinghe:
@gayanper Hopefully now you can proceed with this by resolving the conflicts as I've merged the other PRs. :)  And as @nixel2007 requested, it would be great if you can update the PR description with an image with the next change so that the others can see the styling changes if there are any.

gayanper:
@nixel2007 @NipunaRanasinghe i have update the PR description with a gif.

nixel2007:
@gayanper sorry, i can't get the idea. what's the difference with regular hovers sent from LS?

gayanper:
This is a regular hover, but in this the hyperlinks can be actually clicked and navigate to Files. For example if you click the **Type** link it will open that class file in java editor. if you click the http link it will be opened in the browser etc.

nixel2007:
ah, and you open files via URI parsing and VFS. Gotcha, thanks. Cool!

src/main/java/org/wso2/lsp4intellij/editor/EditorEventManager.java
```diff
@@ -255,9 +266,8 @@ public void mouseMoved(EditorMouseEvent e) {
             return;
         }
         Language language = psiFile.getLanguage();
-        if ((!LanguageDocumentation.INSTANCE.allForLanguage(language).isEmpty() && !language
-                .equals(PlainTextLanguage.INSTANCE)) || (!getIsCtrlDown() && !EditorSettingsExternalizable
-                .getInstance().isShowQuickDocOnMouseOverElement())) {
+        if ((!LanguageDocumentation.INSTANCE.allForLanguage(language).isEmpty() && !isPlainTextLanguage(language))
```
NipunaRanasinghe: Could you please explain the reason behind this cache based new check? 
`language.equals(PlainTextLanguage.INSTANCE)` isn't working for you?

gayanper: Languages like java doesn't inherit thr plain text language. Thats why this new check is added.

NipunaRanasinghe: Is it not working only for java? how about using a `isFileSupported()` check from our `FileUtils` instead, because you are registering the command server denition for `java` file extensions right ? :)

gayanper: @NipunaRanasinghe i think this is a really good suggestion. I have changed the code now to use FileUtils check.

src/main/java/org/wso2/lsp4intellij/editor/EditorEventManager.java
```diff
@@ -297,6 +296,11 @@ public void mouseMoved(EditorMouseEvent e) {
         }
     }
 
+    private boolean isPlainTextLanguageFile(PsiFile file) {
```
NipunaRanasinghe:
```suggestion
    private boolean isSupportedLanguageFile(PsiFile file) {
```

src/main/java/org/wso2/lsp4intellij/editor/EditorEventManager.java
NipunaRanasinghe: @gayanper great :)

src/main/java/org/wso2/lsp4intellij/editor/EditorEventManager.java
NipunaRanasinghe: @gayanper WDYT since now we check for the registered server definitons as well.

gayanper: I think thats a good suggestion. I made the change and pushed it

PR #150, opened 2019-10-14
Add support for file content based connection to LS

gayanper:
This change introduce a new method in LSPExtensionManager so that
implementors can use to control which files should be connected to LS
based on their content or PsiFIle Language.

### DIFF
--- a/src/main/java/org/wso2/lsp4intellij/client/languageserver/wrapper/LanguageServerWrapper.java
+++ b/src/main/java/org/wso2/lsp4intellij/client/languageserver/wrapper/LanguageServerWrapper.java
@@ -90,12 +90,7 @@ import java.io.InputStream;
 import java.io.OutputStream;
 import java.net.URI;
 import java.net.URISyntaxException;
-import java.util.ArrayList;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Map;
-import java.util.Objects;
-import java.util.Set;
+import java.util.*;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ExecutorService;
@@ -111,6 +106,7 @@ import static org.wso2.lsp4intellij.client.languageserver.ServerStatus.STOPPED;
 import static org.wso2.lsp4intellij.requests.Timeout.getTimeout;
 import static org.wso2.lsp4intellij.requests.Timeouts.INIT;
 import static org.wso2.lsp4intellij.requests.Timeouts.SHUTDOWN;
+import static org.wso2.lsp4intellij.utils.ApplicationUtils.computableReadAction;
 import static org.wso2.lsp4intellij.utils.ApplicationUtils.invokeLater;
 import static org.wso2.lsp4intellij.utils.FileUtils.editorToProjectFolderUri;
 import static org.wso2.lsp4intellij.utils.FileUtils.editorToURIString;
@@ -351,7 +347,7 @@ public class LanguageServerWrapper {
                         }
                         // trigger annotators since the this is the first editor which starts the LS
                         // and annotators are executed before LS is boostrap to provide diagnostics
-                        ApplicationUtils.computableReadAction(() -> {
+                        computableReadAction(() -> {
                             DaemonCodeAnalyzer.getInstance(project).restart(
                                     PsiDocumentManager.getInstance(project).getPsiFile(editor.getDocument()));
                             return null;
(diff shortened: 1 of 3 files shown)


src/main/java/org/wso2/lsp4intellij/client/languageserver/wrapper/LanguageServerWrapper.java
```diff
@@ -384,6 +385,17 @@ public void connect(Editor editor) {
...
+        return computableReadAction(() ->
+                Optional.ofNullable(PsiDocumentManager.getInstance(editor.getProject()).getPsiFile(editor.getDocument()))
+                        .map(f ->
+                                Optional.ofNullable(extManager).map(e -> e.isFileContentSupported(f)).orElse(false)
+                        ).orElseGet(() -> {
+                    LOG.debug("PsiFile for editor is null, isFileContentSupported will return false");
+                    return false;
+                }));
+    }
+
```
NipunaRanasinghe: Shall we move this to the `FileUtils` and include this content check inside the `isFileSupported` method? That's the core method which is being used to validate files everywhere and therefore this refactoring would make your life easy ;)

NipunaRanasinghe: Otherwise you'll have to add this check in almost all of those places

NipunaRanasinghe: BTW If you got more works to do, I can do the refactoring on top of this after merging as it is.. WDYT?

gayanper: This only needs to be used as a checkpoint before starting a LS. This is different from the method we have in the FileUtil check. This is a extension point on top of the file extension check. Do you still think we need to move ? If we move we need to take the support for reading the extension manager as well. I feel that will be too much unrelated dependencies to FileUtils. WDYT ?

NipunaRanasinghe: My point is, if we are going to add support for filtering out files based on its content, we might need to do the same guarding in some of our listeners/entry points (e.g. see `VFSListener` and `LSPFileEventManager` implementation). But as you explained it might need some additional works. Anyway we can proceed with this way :smiley: Let get this merged

NipunaRanasinghe:
@gayanper seems like after the last merge, we have a conflict in here as well :/ Shall we resolve?

src/main/java/org/wso2/lsp4intellij/client/languageserver/wrapper/LanguageServerWrapper.java
NipunaRanasinghe: One more question, if the extManager is null, shouldn't we consider as the content is supported like
```
Optional.ofNullable(extManager).map(e -> e.isFileContentSupported(f)).orElse(true)
```

gayanper: I see your point now. I think we should guard the places where we check for file and then send messages for the LS. The current state of this PR can result in invalid file content change messages to a LS. I try to incorporate those changes as well. But isn't that a problem already. I mean if i open a pom file and change it in a bal project, isn't this lib sending a file change message to the bal ls regardless bal ls is not supporting pom files ?

NipunaRanasinghe: @gayanper That's exactly what I was saying :) 

I think currently there's no chance of triggering LS requests for invalid files 
because I'm checking for any language server definition registrations for the file extension/name pattern inside `isFileSupported` check. (See `IntellijLanguageClient.isExtensionSupported(virtualFile)`).
So as per your example `isFileSupported` will return `false` for pom files.

NipunaRanasinghe: > One more question, if the extManager is null, shouldn't we consider as the content is supported as
> 
> ```
> Optional.ofNullable(extManager).map(e -> e.isFileContentSupported(f)).orElse(true)
> ```
> 
> ?

@gayanper WDYT?

gayanper: Have done the changes, it seems the correct place was for this logic in the FileUtils. Thanks @NipunaRanasinghe for pointing out that. Please see if it looks OK now.

gayanper:
Have done the changes, it seems the correct place was for this logic in the FileUtils. Thanks @NipunaRanasinghe for pointing out that. Please see if it looks OK now.

gayanper:
@NipunaRanasinghe once merged i will update the rest of the PRs as well with the master to merge them

PR #151, opened 2019-10-15
fix rename bug, editor does not hasFocus when using rename from menu

sonicning:
lsp-rename-bug

`if (editor.getContentComponent().hasFocus())`

OpenJDK Runtime Environment (build 11.0.3+12-b304.10)
OpenJDK 64-Bit Server VM (build 11.0.3+12-b304.10, mixed mode)

$ ./clangd --version
clangd version 8.0.0 (tags/RELEASE_800/final)

### DIFF
--- a/src/main/java/org/wso2/lsp4intellij/contributors/rename/LSPRenameHandler.java
+++ b/src/main/java/org/wso2/lsp4intellij/contributors/rename/LSPRenameHandler.java
@@ -66,13 +66,13 @@ public class LSPRenameHandler implements RenameHandler {
     public void invoke(@NotNull Project project, Editor editor, PsiFile file, DataContext dataContext) {
         EditorEventManager manager = EditorEventManagerBase.forEditor(editor);
         if (manager != null) {
-            if (editor.getContentComponent().hasFocus()) {
-                LSPPsiElement psiElement = getElementAtOffset(manager,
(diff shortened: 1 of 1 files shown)


src/main/java/org/wso2/lsp4intellij/contributors/rename/LSPRenameHandler.java
```diff
@@ -66,13 +66,13 @@ public void invoke(@NotNull Project project, @NotNull PsiElement[] elements, Dat
...
-                if (psiElement != null) {
-                    doRename(psiElement, editor);
-                }
+            // if (editor.getContentComponent().hasFocus()) {
+            LSPPsiElement psiElement = getElementAtOffset(manager,
+                editor.getCaretModel().getCurrentCaret().getOffset());
+            if (psiElement != null) {
+                doRename(psiElement, editor);
             }
+            // }
```
NipunaRanasinghe: Shall we remove the commented lines here? :)

sonicning: Yes, please. I make it commented just to make the point and make it clear to understand. We should definitely remove the commented lines. :)

NipunaRanasinghe:
Thanks for the fix @sonicning!

PR #161, opened 2019-11-13
[Annotation][QuickFix] the register action logic is after mouse cursor moving on diagnostic annotation, and modify it to createAnnotation logic

MagicRUBICK:
image

### DIFF
--- a/src/main/java/org/wso2/lsp4intellij/contributors/annotator/LSPAnnotator.java
+++ b/src/main/java/org/wso2/lsp4intellij/contributors/annotator/LSPAnnotator.java
@@ -23,10 +23,22 @@ import com.intellij.openapi.editor.Editor;
 import com.intellij.openapi.util.TextRange;
 import com.intellij.openapi.vfs.VirtualFile;
 import com.intellij.psi.PsiFile;
+import org.eclipse.lsp4j.CodeAction;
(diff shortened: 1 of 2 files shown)


NipunaRanasinghe:
@MagicRUBICK This is actually the previous implementation we had and let me explain why we had to move to a more complex diagnostics + code actions implementation.

Basically there are two scenarios which this IntelliJ language client should be able to handle
1. Diagnostics based quick fixes (which you've adressed)
2. Line based quick fixes (which you've missed)

Language server protocol allows providing code actions for a defined range(usually line-based) even without having any diagnostics present on a particular line. Therefore we need to request code actions in each time where the cursor position is moved to a new line. (this is being done by vscode language client as well). On the other hand there might be a considerable performance cost if we always send code actions request on each diagnostic notification coming from the lang server side. So the current implementation is more likely a lazy loading approach. Hope you understand the complications and reasons for the current approach. :)

MagicRUBICK:
Got it, thx. 
Do you know how is the implement in VSCode.

NipunaRanasinghe:
@MagicRUBICK Sorry I haven't had a chance to work on vscode developments so far :/ AFAIR vscode language client is well documented so you might be able to have a clue there :)

NipunaRanasinghe:
Closing the PR as per the above discussion.

PR #165, opened 2019-11-28
[Action] bind the action register with caret changed listener replaci…

MagicRUBICK:


NipunaRanasinghe:
@MagicRUBICK I tested your PR locally and it works great :) Anyway it'll be better to debounce codeaction requests to for better performance specially when the user keeps typing and therefore the cursor is changed rapidly. I'll work on those improvements on top of these changes after merging :) LGTM!

PR #174, opened 2019-12-13
Fix random error in 2019.3

gayanper:
This fix will make sure the file index is not accessed while PSI file
 events are being processed. This is a prohibited action in 2019.3

### DIFF
--- a/src/main/java/org/wso2/lsp4intellij/utils/ApplicationUtils.java
+++ b/src/main/java/org/wso2/lsp4intellij/utils/ApplicationUtils.java
@@ -16,7 +16,11 @@
 package org.wso2.lsp4intellij.utils;
 
 import com.intellij.openapi.application.ApplicationManager;
+import com.intellij.openapi.application.ModalityState;
+import com.intellij.openapi.project.NoAccessDuringPsiEvents;
 import com.intellij.openapi.util.Computable;
+import com.intellij.openapi.util.Condition;
+
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 
@@ -55,5 +59,17 @@ public class ApplicationUtils {
 
     static public <T> T computableWriteAction(Computable<T> computable) {
         return ApplicationManager.getApplication().runWriteAction(computable);
+    }
+
+    static public void invokeAfterPsiEvents(Runnable runnable) {
+        Runnable wrapper = () -> {
+            if(NoAccessDuringPsiEvents.isInsideEventProcessing()) {
+                invokeAfterPsiEvents(runnable);
+            } else {
+                runnable.run();
+            }
+        };
+
+        ApplicationManager.getApplication().invokeLater(wrapper, (Condition<Void>) value -> false);
     }
 }
--- a/src/main/java/org/wso2/lsp4intellij/listeners/LSPFileEventManager.java
+++ b/src/main/java/org/wso2/lsp4intellij/listeners/LSPFileEventManager.java
@@ -82,15 +82,17 @@ class LSPFileEventManager {
             return;
         }
 
-        EditorEventManager manager = EditorEventManagerBase.forUri(uri);
-        if (manager != null) {
-            manager.documentSaved();
-            FileUtils.findProjectsFor(file).forEach(p -> changedConfiguration(uri,
+        ApplicationUtils.invokeAfterPsiEvents(() -> {
+            EditorEventManager manager = EditorEventManagerBase.forUri(uri);
+            if (manager != null) {
+                manager.documentSaved();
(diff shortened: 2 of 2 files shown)


src/main/java/org/wso2/lsp4intellij/utils/ApplicationUtils.java
```diff
@@ -56,4 +60,16 @@ static public void writeAction(Runnable runnable) {
...
         return ApplicationManager.getApplication().runWriteAction(computable);
     }
+
+    static public void invokeAfterPsiEvents(Runnable runnable) {
+        Runnable wrapper = () -> {
+            if(NoAccessDuringPsiEvents.isInsideEventProcessing()) {
+                invokeAfterPsiEvents(runnable);
+            } else {
+                runnable.run();
+            }
```
NipunaRanasinghe:
```java
                Runnable wrapper = () -> {
                    while (NoAccessDuringPsiEvents.isInsideEventProcessing()) {
                        try {
                            Thread.sleep(500);
                        } catch (InterruptedException e) {
                            Thread.currentThread().interrupt();
                        }
                    }
                    runnable.run();
                };
```

IMO it would be better to have intervals between the checks as above. WDYT?

gayanper: I don’t think it’s a good idea to do like this. Because this will cause the event loop to wait stuck on the current event until the psi events are done which might cause a deadlock. The way I handle is by adding the event to tail always to give chance to finish the existing psi events in the event queue

gayanper: I don’t think it’s a good idea to do like this. Because this will cause the event loop to wait stuck on the current event until the psi events are done which might cause a deadlock. The way I handle is by adding the event to tail always to give chance to finish the existing psi events in the event queue

gayanper: I don’t think it’s a good idea to do like this. Because this will cause the event loop to wait stuck on the current event until the psi events are done which might cause a deadlock. The way I handle is by adding the event to tail always to give chance to finish the existing psi events in the event queue

NipunaRanasinghe: @gayanper Forgive me that I'm still not clear about your argument here but what I was worrying is the recursion depth as we are going recursively without any delay. Correct me If I'm wrong :)

gayanper: @NipunaRanasinghe I don’t think we have a recursion here. Because as soon as we push this runnable to invoke later the method will return. If we have a recursion actually this code should not work at all. Because until the current event pops out of the queue the remaining psi events won’t drain. Do you still think it can cause a recursion?

NipunaRanasinghe: @gayanper If I'm not mistaken you are invoking `invokeAfterPsiEvents(Runnable runnable);`  inside its own function body and that's what causes the recursion.

gayanper: @NipunaRanasinghe i think its little bit confusing how this code works, that might be the problem. Let me try to explain how i see it.

The first call comes into invokeAfterPsiEvents and the invocation is wrapped inside the new runnable. That runnable is pushed into the application event queue (AEQ). As soon as it it done the invokeAfterPsiEvents returns. So when the time comes to this runnable to run, it will check if still there are pending PSI events. If so we again call the invokeAfterPsiEvents with the orginal runnable instance we got. Now from the above we know that invokeAfterPsiEvents returns as soon as it pushes to AEQ. So i'm some what sure that there cannot be a recursion senario because we are using the second runnable wrapper and the AEQ. Do you still think there is a recursion ? if so please explain the steps how the recursion can occur because i might be still missing the view point you are having.

gayanper:
@NipunaRanasinghe are we going to merge this soon ? I think the recursion concern was cleared out right ? I have been running with this fix for several weeks now and i can say i didn't find any side effects.

src/main/java/org/wso2/lsp4intellij/utils/ApplicationUtils.java
NipunaRanasinghe: Thanks for your explanation and I also think your concern is valid. It was just the implementation being somewhat tricky to understand at once, but now I also thinks that this should work as above. :)

NipunaRanasinghe:
@gayanper sorry for late response but I think you were right. I've also been testing this fix locally for few weeks and didn't notice anything strange. LGTM :)

PR #176, opened 2020-01-02
Add support for scoping symbols

gayanper:
Add support for scoping the symbol results using IntelliJ default scope options.

Now the open symbol popup's default scoping options can be applied to LSP results as well.


gayanper:
Seems like the copyright headers needs to update on all files right @NipunaRanasinghe ?

NipunaRanasinghe:
@gayanper Yeah I'll send a separate PR to fix it. :+1:  LGTM

PR #187, opened 2020-05-19
Fixed compatibility issues with upcoming IJ release

gayanper:
Fix for the following compatibility issue
```
Compatibility problems (1): 
    #Access to unresolved field com.intellij.icons.AllIcons.Nodes.DataSource : Icon
        Method org.wso2.lsp4intellij.contributors.icon.LSPDefaultIconProvider.getCompletionIcon(org.eclipse.lsp4j.CompletionItemKind kind) : javax.swing.Icon contains a *getstatic* instruction referencing an unresolved field com.intellij.icons.AllIcons.Nodes.DataSource : javax.swing.Icon. This can lead to **NoSuchFieldError** exception at runtime.
```
When verified against: IC-201.6668.121

Issue #184, opened 2020-03-26
Add support for Intellij 2020.1


gayanper:
I tried testing my STS4 plugin with 2020.1 seems like even though the language server process is started, the messages are not sent to the language server. The language server icon stays yellow forever.

Since this is the upcoming release of IntelliJ we might need to look at what going on soon and apply a fix for it.

NipunaRanasinghe:
Yeah we should check on that. One possible reason might be registering the client as an application component (application components are not recommended in the dynamic plugin concept which will be available with the 2020.1 release). Thanks for bringing this up.

gayanper:
I checked the code and it seems it was a issue in the plugin that i have been working on. But it seems that there are some generic type changes in latest IDEA platform which causes compilation errors. But i think that won't be a issue until we try to compile the plugin against 2020.1. Shall we close this issue @NipunaRanasinghe ?

NipunaRanasinghe:
@gayanper Yeah when I tried to compile against IDEA 2020.1, compilation was failed due to few deprecated icons and type changes in `LSPSymbolContributor`, but the library seems to be running fine so far. Anyway we better keep this opened until we fix the above compilation issues, IMO.

gayanper:
@NipunaRanasinghe following issues stops me from submitting my spring plugin into marketplace

```
Compatibility problems (1): 
    #Access to unresolved field com.intellij.icons.AllIcons.Nodes.DataSource : Icon
        Method org.wso2.lsp4intellij.contributors.icon.LSPDefaultIconProvider.getCompletionIcon(org.eclipse.lsp4j.CompletionItemKind kind) : javax.swing.Icon contains a *getstatic* instruction referencing an unresolved field com.intellij.icons.AllIcons.Nodes.DataSource : javax.swing.Icon. This can lead to **NoSuchFieldError** exception at runtime.
Deprecated API usages (23): 
    #Deprecated method com.intellij.openapi.components.BaseComponent.initComponent() is overridden
        Deprecated method com.intellij.openapi.components.BaseComponent.initComponent() : void is overridden in class org.wso2.lsp4intellij.IntellijLanguageClient
    #Deprecated method com.intellij.openapi.wm.StatusBarWidget.getPresentation(StatusBarWidget.PlatformType) is overridden
[...]

PR #188, opened 2020-05-19
Update for IDEA 2020.1

carymrobbins:
This updates the build to target IDEA 2020.1. I'm not sure if we want to leave the `build.gradle` as-is or update it to 2020.1 as I have in this PR. There's certainly value in updating it to verify that the build actually compiles with a recent version of IntelliJ. If we want to ensure cross-IntelliJ compatiblity, we should probably update the CI job to deal with that in some way.

I've titled this PR as `WIP` as the `FileUtilsTest` is failing for me. However, HEAD doesn't build for me for other reasons, so if it passes on CI I'd be ok with moving forward.

```
> Task :test

org.wso2.lsp4intellij.utils.FileUtilsTest > initializationError FAILED
    org.objenesis.ObjenesisException
        Caused by: java.lang.reflect.InvocationTargetException
            Caused by: java.lang.IllegalAccessError

9 tests completed, 1 failed
```

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -15,7 +15,7 @@ group = "org.wso2.lsp4intellij"
 version = gitVersionCalculator.calculateVersion("v")
 
 intellij {
-    version '2017.3'
+    version '2020.1'
     type 'IC'
     updateSinceUntilBuild false
 }
--- a/src/main/java/org/wso2/lsp4intellij/contributors/psi/LSPPsiElement.java
+++ b/src/main/java/org/wso2/lsp4intellij/contributors/psi/LSPPsiElement.java
@@ -358,6 +358,7 @@ public class LSPPsiElement implements PsiNameIdentifierOwner, NavigatablePsiElem
      * @throws IncorrectOperationException if the modification is not supported or not possible for some reason.
      * @deprecated not all PSI implementations implement this method correctly.
      */
+    @Deprecated
     public void checkAdd(PsiElement element) {
         throw new IncorrectOperationException();
     }
@@ -417,6 +418,7 @@ public class LSPPsiElement implements PsiNameIdentifierOwner, NavigatablePsiElem
      * @throws IncorrectOperationException if the modification is not supported or not possible for some reason.
      * @deprecated not all PSI implementations implement this method correctly.
      */
+    @Deprecated
     public void checkDelete() {
         throw new IncorrectOperationException();
     }
--- a/src/main/java/org/wso2/lsp4intellij/contributors/symbol/LSPSymbolContributor.java
+++ b/src/main/java/org/wso2/lsp4intellij/contributors/symbol/LSPSymbolContributor.java
@@ -32,10 +32,10 @@ import org.jetbrains.annotations.Nullable;
  */
 public class LSPSymbolContributor implements ChooseByNameContributorEx {
 
-    private WorkspaceSymbolProvider workspaceSymbolProvider = new WorkspaceSymbolProvider();
+    private final WorkspaceSymbolProvider workspaceSymbolProvider = new WorkspaceSymbolProvider();
 
     @Override
-    public void processNames(@NotNull Processor<String> processor, @NotNull GlobalSearchScope globalSearchScope, @Nullable IdFilter idFilter) {
+    public void processNames(@NotNull Processor<? super String> processor, @NotNull GlobalSearchScope globalSearchScope, @Nullable IdFilter idFilter) {
         workspaceSymbolProvider.workspaceSymbols("", globalSearchScope.getProject()).stream()
             .filter(ni -> globalSearchScope.accept(ni.getFile()))
             .map(NavigationItem::getName)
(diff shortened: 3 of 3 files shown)


carymrobbins:
Just noticed that this intersects with PR https://github.com/ballerina-platform/lsp4intellij/pull/187 which was just opened a few hours ago. However, this PR fixes additional issues that were causing compilation errors when building against 2020.1.

NipunaRanasinghe:
@carymrobbins First of all, really appreciate your effort on this. I've been busy with few other works and just had a chance to continue this work.
Regarding the idea build version, we've been keeping the minimum compatible version in order to make sure that all our changes + compatibility fixes are doesn't affect the oldest version of the compatible range (and if not, we can bump the minimum compatible version).  But as you mentioned, we also have to validate against the latest stable version as well. So I'm okay with either approach until we figure out a way validate the build against multiple IDEA versions.
(And I can see that with your changes on `LSPSymbolContributor`, the minimum compatible version won't be 2017.3 anymore. So I'll update the REAME with a compatibility matrix with the next patch release.)

Regarding the test failure, make sure you are using  jdk8. Anyway I guess we are all good to merge after resolving the conflicts which caused by #187.

nixel2007:
> Regarding the test failure, make sure you are using jdk8

AFAIR intellij platform 2020.1+ uses jdk11 by default. should we stick with jdk8?

NipunaRanasinghe:
> > Regarding the test failure, make sure you are using jdk8
> 
> AFAIR intellij platform 2020.1+ uses jdk11 by default. should we stick with jdk8?

Yeah AFAIK default JBRE version for IntelliJ 2020.x versions will be jdk11. I just meant that until we verify that the library is compatible with higher jdk versions (and fix if not), we have to use the jdk version we've been using. BTW have you tried to build with higher jdk versions and encountered any issues?

nixel2007:
> BTW have you tried to build with higher jdk versions and encountered any issues?

build - no. sorry. but test latest versions in 2020.1 with bundled bsl ls without problem.

carymrobbins:
Mind if I rebase to fix the merge conflicts?

carymrobbins:
The conflicts were only because of a change I made to `LSPDefaultIconProvider.java`, which I've reverted, so we should be all good now.

NipunaRanasinghe:
> The conflicts were only because of a change I made to `LSPDefaultIconProvider.java`, which I've reverted, so we should be all good now.

perfect! Thanks @carymrobbins.

carymrobbins:
> Could you please also verify the oldest IDEA version we can support after the change, so that we can update the compatibility details?

I'm not really sure what the best way to go about this. From what I can tell, the method signatures changed in this commit - https://github.com/JetBrains/intellij-community/commit/0ac6e724c94bb94ccba51a378d89a4f86af2b8d5

That commit appears in these tags -

```sh-session
% cd intellij-community
% git fetch --tags
% git tag --contains 0ac6e724c94bb94ccba51a378d89a4f86af2b8d5
idea/201.4515.24
idea/201.4865.12
idea/201.5259.13
idea/201.5616.10
idea/201.5985
idea/201.5985.32
idea/201.6073.9
idea/201.6251.22
idea/201.6251.9
idea/201.6487.11
idea/201.6668.113
idea/201.6668.121
idea/201.6668.13
idea/201.6668.60
idea/201.7223.18
idea/201.7223.58
idea/201.7223.91
idea/201.7846.29
idea/201.7846.6
```

So I'm guessing that the minimum compatible IntelliJ would be the `201.4515.24` build, which I believe is some variant of the `2020.1` release.

NipunaRanasinghe:
@carymrobbins I also tried to build with IntelliJ 2019.3 and it failed, which means that the **minimum supported IDEA version with the API change will be 2020.1** itself. @gayanper WDYT?

gayanper:
@NipunaRanasinghe well this will cause all the plugins that depend on this library to have their minimum version as 2020.1, which is fine by me, but probably not by many. Of course for my plugin I don't see why not upgrade to the latest CE version. But for plugins that are targetting the Ultimate version would be a problem. It's always good to keep the compatibility to at least to version (latest version) - 1 IMO.

carymrobbins:
One thing you could do is have some sort of cross build that targets multiple IDEA versions. You would need to have conditional source directories that are only used when a particular version of IDEA is being built against.

However, I wonder if the plugin will still "just work" when used on an older IDEA version? Unless I'm missing something, the only breaking change was the generic type params, which are subject to erasure and shouldn't produce incompatible JVM bytecode for earlier versions.

gayanper:
> One thing you could do is have some sort of cross build that targets multiple IDEA versions. You would need to have conditional source directories that are only used when a particular version of IDEA is being built against.
> 
> However, I wonder if the plugin will still "just work" when used on an older IDEA version? Unless I'm missing something, the only breaking change was the generic type params, which are subject to erasure and shouldn't produce incompatible JVM bytecode for earlier versions.

I think we should have kept them compatible for 2019.1 as well without changing the signatures as it was before, because the same type erasure applies there for 2020.1 as well. Keeping the compilation against 2019.1 is a good balance as a library.

nixel2007:
> 2019.1

but latest `CE - 1 version` is 2019.3...

gayanper:
> > 2019.1
> 
> but latest `CE - 1 version` is 2019.3...
True, what i wanted to say is we need to support the previouse major version (including all 3 minor versions).

NipunaRanasinghe:
@carymrobbins @nixel2007 please refer the discussion in https://github.com/ballerina-platform/lsp4intellij/issues/184, where I also think that it would be better keep the compatibility at least for the previous set of "major"(2019.x) IntelliJ releases.

PR #196, opened 2020-05-26
Make LSPAnnotator more extensible

carymrobbins:
I'm not sure if this is the right way to go about this, but we likely want some way of customizing the highlighting of LSPAnnotator. Extending LSPAnnotator and customizing its behavior with this new method, `modifyCreatedAnnotation()`, seems to make this work quite easily. Though, again, I'm open to a better way of accomplishing this.

For context, here is how I'm using this -

```scala
class HaskForceLSPAnnotator extends LSPAnnotator {

  override def modifyCreatedAnnotation(annotation: Annotation): Unit = {
    GhcMessageType.of(annotation.getMessage).foreach {
      case GhcMessageType.NotInScope =>
        annotation.setHighlightType(ProblemHighlightType.LIKE_UNKNOWN_SYMBOL)
      case GhcMessageType.CouldNotFindModule =>
        annotation.setHighlightType(ProblemHighlightType.LIKE_UNKNOWN_SYMBOL)
    }
  }
}
```

### DIFF
--- a/src/main/java/org/wso2/lsp4intellij/contributors/annotator/LSPAnnotator.java
+++ b/src/main/java/org/wso2/lsp4intellij/contributors/annotator/LSPAnnotator.java
@@ -36,7 +36,7 @@ import java.util.ArrayList;
 import java.util.ConcurrentModificationException;
 import java.util.List;
 
-public class LSPAnnotator extends ExternalAnnotator {
+public class LSPAnnotator extends ExternalAnnotator<Object, Object> {
 
     private static final Logger LOG = Logger.getInstance(LSPAnnotator.class);
     private static final Object RESULT = new Object();
@@ -110,6 +110,7 @@ public class LSPAnnotator extends ExternalAnnotator {
             return;
         }
         annotations.forEach(annotation -> {
+            // TODO: Use 'newAnnotation'; 'createAnnotation' is deprecated.
             Annotation anon = holder.createAnnotation(annotation.getSeverity(),
                     new TextRange(annotation.getStartOffset(), annotation.getEndOffset()), annotation.getMessage());
 
@@ -120,11 +121,25 @@ public class LSPAnnotator extends ExternalAnnotator {
         });
     }
 
-    /**
-     * This can be overridden to modify an annotation after it has been created.
-     * Useful for customizing highlighting beyond what LSP supports.
-     */
-    protected void modifyCreatedAnnotation(@SuppressWarnings("unused") Annotation annotation) {
+    @Nullable
+    protected Annotation createAnnotation(Editor editor, AnnotationHolder holder, Diagnostic diagnostic) {
+        final int start = DocumentUtils.LSPPosToOffset(editor, diagnostic.getRange().getStart());
+        final int end = DocumentUtils.LSPPosToOffset(editor, diagnostic.getRange().getEnd());
+        if (start >= end) {
+            return null;
+        }
+        final TextRange textRange = new TextRange(start, end);
+        switch (diagnostic.getSeverity()) {
+            // TODO: Use 'newAnnotation'; 'create*Annotation' methods are deprecated.
+            case Error:
+                return holder.createErrorAnnotation(textRange, diagnostic.getMessage());
+            case Warning:
+                return holder.createWarningAnnotation(textRange, diagnostic.getMessage());
+            case Information:
+                return holder.createInfoAnnotation(textRange, diagnostic.getMessage());
+            default:
+                return holder.createWeakWarningAnnotation(textRange, diagnostic.getMessage());
+        }
     }
 
     private void createAnnotations(AnnotationHolder holder, EditorEventManager eventManager) {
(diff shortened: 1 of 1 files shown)


carymrobbins:
@NipunaRanasinghe Could you expand on this?

> You can write your own util method for createAnnotations() and use it by overriding apply() (In that case, you dont have to do any changes in the library, as the apply() is already public)

It's not clear how I can make this work without essentially reimplementing LSPAnnotator. Maybe I'm missing something obvious though.

NipunaRanasinghe:
> @NipunaRanasinghe Could you expand on this?
> 
> > You can write your own util method for createAnnotations() and use it by overriding apply() (In that case, you dont have to do any changes in the library, as the apply() is already public)
> 
> It's not clear how I can make this work without essentially reimplementing LSPAnnotator. Maybe I'm missing something obvious though.

@carymrobbins Sorry if my previous answer was not clear enough. What I meant was,

1. You can write your own util method for createAnnotations() and use it by overriding apply() (In that case, you dont have to do any changes in the library, as the apply() is already public).

For your context, it might look like this.
```scala
class HaskForceLSPAnnotator extends LSPAnnotator {

  override def apply(file: PsiFile, annotationResult: Object, holder: AnnotationHolder): Unit = {
      // you can copy paste the content from the super method and replace the updateAnnotation()
      // invocation with your custom util method.
  }
}
```

2. We can think of changing the access modifiers of the utils functions into `protected`, which makes  the default annotator fully extendable.

For your context, it might look like this.
```scala
class HaskForceLSPAnnotator extends LSPAnnotator {

   override def createAnnotations(file: PsiFile, annotationResult: Object, holder: AnnotationHolder): Unit = {
     // Here you can copy paste the content of the super method and add you customizations at the 
     // end. To be able to override this util method, we can change the access modifier as
     // private -> protected in LSPAnnotator class.
  }
}
```

carymrobbins:
I see. I guess I was trying to come up with something that didn't require copy paste. I think for this stuff to be useful to plugin authors we need some way of making things like this extendable.

NipunaRanasinghe:
> I see. I guess I was trying to come up with something that didn't require copy paste. I think for this stuff to be useful to plugin authors we need some way of making things like this extendable.

Yes exactly. I do see that you have a good enough solution for your use case but the problem is with the scalability. Just as an example, if someone wants to filter out some annotations based on a property, probably he won't be able to use the method you are proposing. That's why I think that the better approach would be making the methods (including private utils) "overridable" so that the users(plugin authors) will have more control over customizing the flow for their language specific requirements. @gayanper WDYT?

gayanper:
I think we need to see how often one want to change the default presentation and how often some one want to totally ignore an annotation. I see the first one as most common and i do like letting the consumer to just change what they want with some default template methods. So +1 for @carymrobbins solution. May be in second case some one should totally write the content of apply method or else we could provide support with a predicate API when mapping the annotations

carymrobbins:
Maybe this would be a better way to do it -

```java
@Nullable
protected Annotation createAnnotation(Editor editor, AnnotationHolder holder, Diagnostic diagnostic) {
    final int start = DocumentUtils.LSPPosToOffset(editor, diagnostic.getRange().getStart());
    final int end = DocumentUtils.LSPPosToOffset(editor, diagnostic.getRange().getEnd());
    if (start >= end) {
        return null;
    }
    final TextRange textRange = new TextRange(start, end);
    switch (diagnostic.getSeverity()) {
        case Error:
            return holder.createErrorAnnotation(textRange, diagnostic.getMessage());
        case Warning:
            return holder.createWarningAnnotation(textRange, diagnostic.getMessage());
        case Information:
            return holder.createInfoAnnotation(textRange, diagnostic.getMessage());
        default:
            return holder.createWeakWarningAnnotation(textRange, diagnostic.getMessage());
    }
}
private void createAnnotations(AnnotationHolder holder, EditorEventManager eventManager) {
    final List<Diagnostic> diagnostics = eventManager.getDiagnostics();
    final Editor editor = eventManager.editor;
    List<Annotation> annotations = new ArrayList<>();
    diagnostics.forEach(d -> {
        Annotation annotation = createAnnotation(editor, holder, d);
        if (annotation != null) {
            annotations.add(annotation);
        }
    });
    eventManager.setAnnotations(annotations);
    eventManager.setAnonHolder(holder);
}
```

Then I could do this

```scala
class HaskForceLSPAnnotator extends LSPAnnotator {

  override protected def createAnnotation(editor: Editor, holder: AnnotationHolder, diagnostic: Diagnostic): Annotation = {
    val annotation = super.createAnnotation(editor, holder, diagnostic)
    GhcMessageType.of(annotation.getMessage).foreach {
      case GhcMessageType.NotInScope =>
        annotation.setHighlightType(ProblemHighlightType.LIKE_UNKNOWN_SYMBOL)
      case GhcMessageType.CouldNotFindModule =>
        annotation.setHighlightType(ProblemHighlightType.LIKE_UNKNOWN_SYMBOL)
    }
    annotation
  }
}
```

This seems to be a little less intrusive while avoiding pushing the burden on plugin authors to re-implement existing logic.

carymrobbins:
I just pushed a commit with this change. I can revert if we think the other is better, but I think this one is a bit more flexible. It also covers the case of wanting to filter out annotations (you can return `null` from your overridden `createAnnotation` to do this.

carymrobbins:
Also, it seems that this technique for creating annotations is deprecated (I've added relevant TODOs). The "new" way is to use `AnnotationHolder.newAnnotation` and call `AnnotationBuilder.create` which returns `void`, so it seems we can't really rely on holding onto `Annotation` values for much longer. Maybe we should just be using `AnnotationBuilder` values instead of `Annotation`?

carymrobbins:
> Also, it seems that this technique for creating annotations is deprecated (I've added relevant TODOs). The "new" way is to use `AnnotationHolder.newAnnotation` and call `AnnotationBuilder.create` which returns `void`, so it seems we can't really rely on holding onto `Annotation` values for much longer. Maybe we should just be using `AnnotationBuilder` values instead of `Annotation`?

I went ahead and cut a ticket (https://github.com/ballerina-platform/lsp4intellij/issues/200) for this so it can be handled separately. Seems there will need to be some discussion around it.

NipunaRanasinghe:
> > Also, it seems that this technique for creating annotations is deprecated (I've added relevant TODOs). The "new" way is to use `AnnotationHolder.newAnnotation` and call `AnnotationBuilder.create` which returns `void`, so it seems we can't really rely on holding onto `Annotation` values for much longer. Maybe we should just be using `AnnotationBuilder` values instead of `Annotation`?
> 
> I went ahead and cut a ticket (#200) for this so it can be handled separately. Seems there will need to be some discussion around it.

Yeah that's better. We can separately think of getting rid of the deprecated annotator APIs.

PR #197, opened 2020-05-26
Fix NPEs in DefaultRequestManager

carymrobbins:
Fixes more NPEs due to nullable Boolean values from lsp4j.

Side note - is assuming `false` as the default ok in these cases? Or is this actually a bug with us not initializing the fields correctly?

### DIFF
--- a/src/main/java/org/wso2/lsp4intellij/client/languageserver/requestmanager/DefaultRequestManager.java
+++ b/src/main/java/org/wso2/lsp4intellij/client/languageserver/requestmanager/DefaultRequestManager.java
@@ -414,9 +414,9 @@ public class DefaultRequestManager implements RequestManager {
         if (checkStatus()) {
             try {
                 return (
-                    Optional.of(serverCapabilities.getHoverProvider()).orElse(false)
-                ) ? textDocumentService.hover(params)
-                  : null;
+                    Optional.of(serverCapabilities.getHoverProvider()).orElse(false)) ?
+                        textDocumentService.hover(params) : null;
+               
             } catch (Exception e) {
                 crashed(e);
                 return null;
--- a/src/main/java/org/wso2/lsp4intellij/client/languageserver/requestmanager/DefaultRequestManager.java
+++ b/src/main/java/org/wso2/lsp4intellij/client/languageserver/requestmanager/DefaultRequestManager.java
@@ -398,12 +398,9 @@ public class DefaultRequestManager implements RequestManager {
     public CompletableFuture<CompletionItem> resolveCompletionItem(CompletionItem unresolved) {
         if (checkStatus()) {
             try {
-                return (
-                    Optional.of(serverCapabilities.getCompletionProvider())
-                        .map(x -> x.getResolveProvider())
-                        .orElse(false)
-                ) ? textDocumentService.resolveCompletionItem(unresolved)
-                  : null;
+                return (Optional.of(serverCapabilities.getCompletionProvider())
+                    .map(x -> x.getResolveProvider()).orElse(false)) ?
+                    textDocumentService.resolveCompletionItem(unresolved) : null;
             } catch (Exception e) {
                 crashed(e);
                 return null;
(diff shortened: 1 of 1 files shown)


src/main/java/org/wso2/lsp4intellij/client/languageserver/requestmanager/DefaultRequestManager.java
```diff
@@ -397,8 +398,12 @@ public void didClose(DidCloseTextDocumentParams params) {
...
         if (checkStatus()) {
             try {
-                return (serverCapabilities.getCompletionProvider() != null && serverCapabilities.getCompletionProvider()
-                        .getResolveProvider()) ? textDocumentService.resolveCompletionItem(unresolved) : null;
+                return (
+                    Optional.of(serverCapabilities.getCompletionProvider())
+                        .map(x -> x.getResolveProvider())
+                        .orElse(false)
+                ) ? textDocumentService.resolveCompletionItem(unresolved)
+                  : null;
```
NipunaRanasinghe, line 406:
```suggestion
                return (Optional.of(serverCapabilities.getCompletionProvider())
                    .map(x -> x.getResolveProvider()).orElse(false)) ?
                    textDocumentService.resolveCompletionItem(unresolved) : null;
```

src/main/java/org/wso2/lsp4intellij/client/languageserver/requestmanager/DefaultRequestManager.java
```diff
@@ -411,7 +416,10 @@ public void didClose(DidCloseTextDocumentParams params) {
     public CompletableFuture<Hover> hover(TextDocumentPositionParams params) {
         if (checkStatus()) {
             try {
-                return (serverCapabilities.getHoverProvider()) ? textDocumentService.hover(params) : null;
+                return (
+                    Optional.of(serverCapabilities.getHoverProvider()).orElse(false)
+                ) ? textDocumentService.hover(params)
+                  : null;
```
NipunaRanasinghe, line 422:
```suggestion
                return (
                    Optional.of(serverCapabilities.getHoverProvider()).orElse(false)) ?
                        textDocumentService.hover(params) : null;
               
```

src/main/java/org/wso2/lsp4intellij/client/languageserver/requestmanager/DefaultRequestManager.java
```diff
@@ -397,8 +398,12 @@ public void didClose(DidCloseTextDocumentParams params) {
     public CompletableFuture<CompletionItem> resolveCompletionItem(CompletionItem unresolved) {
         if (checkStatus()) {
             try {
-                return (serverCapabilities.getCompletionProvider() != null && serverCapabilities.getCompletionProvider()
-                        .getResolveProvider()) ? textDocumentService.resolveCompletionItem(unresolved) : null;
+                return (
+                    Optional.of(serverCapabilities.getCompletionProvider())
+                        .map(x -> x.getResolveProvider())
```
nixel2007, line 403: `.map(CompletionProvider::getResolveProvider)`

and so on... @NipunaRanasinghe WDYT?

carymrobbins: I intentionally used a lambda instead as it reads a bit easier to me. Doesn't seem valuable to make things mire verbose than necessary IMHO.

nixel2007: `::` syntax also provides info about type of `x`, but I understand your point.

NipunaRanasinghe: > `.map(CompletionProvider::getResolveProvider)`
> 
> and so on... @NipunaRanasinghe WDYT?

@nixel2007 @matthewmichihara I'm being 0 here as the 2 approaches don't have a performance different so it's just a matter of readability + personal preference. :)

NipunaRanasinghe:
@carymrobbins Shall we add the requested changes so that we can go ahead and merge this?

carymrobbins:
@NipunaRanasinghe Yes, sorry for the delay. I've went ahead and auto-applied your suggestions.

NipunaRanasinghe:
> @NipunaRanasinghe Yes, sorry for the delay. I've went ahead and auto-applied your suggestions.

Cool!

PR #199, opened 2020-05-30
Add support for passing in symbol pattern as a query

gayanper:
This will add support to pass in what is ready typed in the Symbols Popup filter as the query to symbols request.

This is valuable when LS will limit the number of symbols that are returned for performance reasons. For example STS4 LS will only return 50 symbols in a single invocation. So you will need to narrow down the search query to get the symbols you need. 

This change will help such LS implementations.


gayanper:
@NipunaRanasinghe the solution I have used it something that as already mentioned in the https://jetbrains-platform.slack.com/archives/C5U8BM1MK/p1588682056402200?thread_ts=1588578964.365900&cid=C5U8BM1MK

PR #201, opened 2020-06-04
upgrade with backward compat to lsp4j 0.9.0

gayanper:
Upgrade to latest lsp4j 0.9.0

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -21,7 +21,7 @@ intellij {
 }
 
 dependencies {
-    compile group: 'org.eclipse.lsp4j', name: 'org.eclipse.lsp4j', version: '0.8.1'
+    compile group: 'org.eclipse.lsp4j', name: 'org.eclipse.lsp4j', version: '0.9.0'
     compile group: 'com.vladsch.flexmark', name: 'flexmark', version: '0.34.58'
     compile group: 'org.apache.commons', name: 'commons-lang3', version: '3.4'
     testCompile group: 'junit', name: 'junit', version: '4.11'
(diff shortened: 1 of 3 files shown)


NipunaRanasinghe:
@gayanper just to be sure, I think it would be okay to migrate the lsp4j version(without having a vote like last time ;)), as this time we ensure the backward compatibility for 0.8.x servers as well right?

nixel2007:
Do we really need to preserve backward comapibility with lsp4j? it's just a wrapper over language server protocol, lsp4intellij doesn't call actual code of language servers, all of this is hidden behind jsop-rpc and lsp4j itself.

nixel2007:
bsl ls migrated to lsp4j 0.9.0 but it works perfectly with lsp4intellij with lsp4j 0.8.0.

gayanper:
I found few methods that doesn’t support when running with lsp4j 0.9.0 like the definitions endpoint which has new specific parameters. @Nipuna yes i have kept backward compatibility for methods that we have integrated in to IJ. But I directly changed the signatures of implementation and typeDefinition endpoints

NipunaRanasinghe:
> Do we really need to preserve backward comapibility with lsp4j? it's just a wrapper over language server protocol, lsp4intellij doesn't call actual code of language servers, all of this is hidden behind jsop-rpc and lsp4j itself.

@nixel2007  0.9.0 release has introduced few method signature changes, which should break the compatibility between 0.8.x and .0.9.x servers/clients.
Its always not possible to do that with new lsp4j versions having backward incompatible changes. But what @gayanper has done is the right thing as we can make sure that aleast 0.8.x and 0.9.x servers will be compatible for now.

NipunaRanasinghe:
> @nipuna yes i have kept backward compatibility for methods that we have integrated in to IJ. But I directly changed the signatures of implementation and typeDefinition endpoints

great! And I believe the reason why you directly changed the signatures for `implementation` and `typeDefinition` operations is we don't support them yet. So that would be fine I guess.

PR #225, opened 2020-11-20
Add tabsize to FormattingOptions

### LINKED ISSUE #208: Formatting request doesn't provide indent amount/tabsize
**Description:**
Formatting request doesn't provide indent amount/tabsize when making a request, which is required by the spec.

See: https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocument_formatting

**Suggested Labels:**
Type/Bug

**Suggested Assignees:**

**Affected Product Version:**
0.94.2

**OS, DB, other environment details and versions:**    
Windows 10, IntelliJ 2020.1.1, https://github.com/Denu8thell/intellij-prisma/tree/prisma-language-server-integration

**Steps to reproduce:**
Pull, compile, and run the above plugin, attempt to format, which does nothing (the language server fails to format, and fails silently due to a different bug).

BinaryFissionGames:
Fixes #208 

Use the projects tabsize and sets the corresponding property on the FormattingOptions object before making a reformat request.

### DIFF
--- a/src/main/java/org/wso2/lsp4intellij/editor/EditorEventManager.java
+++ b/src/main/java/org/wso2/lsp4intellij/editor/EditorEventManager.java
@@ -687,6 +687,7 @@ public class EditorEventManager {
             params.setTextDocument(identifier);
             FormattingOptions options = new FormattingOptions();
             options.setTabSize(DocumentUtils.getTabSize(editor));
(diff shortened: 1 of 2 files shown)


nixel2007:
What about insertSpaces option?

BinaryFissionGames:
Good point, completely glossed over that. Should be good to go now.

NipunaRanasinghe (approved):
LGTM

PR #231, opened 2020-12-27
fixes removal of listeners in case another editor w/same uri is closed

### LINKED ISSUE #229: editor listeners are removed after references toolwindow is closed
listeners are removed if another editor instance with the same uri is closed;
happens after closing references tool window/commit dialog.

i guess it affects closing a split editor, too.

swissiety:
happens e.g. in preview editors like in references or commit dialog;

limits usage of lsp features to editors which are shown in tabs -> EditorKind.MAIN_EDITOR

closes #229

### DIFF
--- a/src/main/java/org/wso2/lsp4intellij/listeners/LSPEditorListener.java
+++ b/src/main/java/org/wso2/lsp4intellij/listeners/LSPEditorListener.java
@@ -15,18 +15,24 @@
  */
 package org.wso2.lsp4intellij.listeners;
 
-import com.intellij.openapi.editor.event.EditorFactoryEvent;
-import com.intellij.openapi.editor.event.EditorFactoryListener;
-import org.jetbrains.annotations.NotNull;
-import org.wso2.lsp4intellij.IntellijLanguageClient;
+        import com.intellij.openapi.components.ServiceManager;
+        import com.intellij.openapi.editor.EditorKind;
+        import com.intellij.openapi.editor.event.EditorFactoryEvent;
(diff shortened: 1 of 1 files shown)


Trias:
hm. i guess the real fix would be to not use a 1 to 1 file to editor relationship. this would fix a whole bunch of issues. unfortunately breaking for everyone using the current api..

 @maintainers: any guidelines for this? i'll have some time the next days for fixing this.

swissiety:
i'll totally agree with you on that matter - it would be a (quick) fix for the current model.

Maybe we can use an Editor object as identifier? But idk when/if editor objects gets disposed/replaced. Maybe there is something better.

swissiety:
*does not work for a split editor duplicate as it just checks whether its a "main editor"

NipunaRanasinghe:
> hm. i guess the real fix would be to not use a 1 to 1 file to editor relationship. this would fix a whole bunch of issues. unfortunately breaking for everyone using the current api..
> 
> @maintainers: any guidelines for this? i'll have some time the next days for fixing this.

@Trias Like you said, the ideal fix would be improving the implementation to support 1-to-many relationship between files and editors. This requirement was already captured sometimes back and has been dragging since it requires some refactoring + proper testing. So I'd be more than grateful if you will have some spare time to work on this :)

Trias:
@NipunaRanasinghe see https://github.com/ballerina-platform/lsp4intellij/pull/241 :)

NipunaRanasinghe (approved):
Lets get this merged

PR #245, opened 2021-02-17
add Deprecations

### LINKED ISSUE #200: Annotation Deprecation
There are multiple usages of the deprecated annotation API, namely the use of `AnnotationHolder.create{Info,Warning,Error}Annotation()` which returns an `Annotation`. These should be replaced with `AnnotationHolder.newAnnotation()` which returns an `AnnotationBuilder`. With the latter, you never get handed an `Annotation`; instead, you build an annotation via the methods on `AnnotationBuilder` and finally call `.create()` which returns `void()` to display the annotation.

Currently, we have two classes which use these deprecated methods, `LSPAnnotator` and `EditorEventManager`.

I think the following is a start towards updating `LSPAnnotator.createAnnotations()`.

```java
    @Nullable
    protected AnnotationBuilder createAnnotationBuilder(Editor editor, AnnotationHolder holder, Diagnostic diagnostic) {
        final int start = DocumentUtils.LSPPosToOffset(editor, diagnostic.getRange().getStart());
        final int end = DocumentUtils.LSPPosToOffset(editor, diagnostic.getRange().getEnd());
        if (start >= end) {
            return null;
        }
        final TextRange textRange = new TextRange(start, end);
        return
          holder
            .newAnnotation(getHighlightSeverity(diagnostic), diagnostic.getMessage())
            .range(textRange);
    }

    private HighlightSeverity getHighlightSeverity(Diagnostic diagnostic) {
        switch (diagnostic.getSeverity()) {
            case Error:
                return HighlightSeverity.ERROR;
            case Warning:
                return HighlightSeverity.WARNING;
            case Information:
                return HighlightSeverity.INFORMATION;
            default:
                return HighlightSeverity.WEAK_WARNING;
        }
    }

    private void createAnnotations(AnnotationHolder holder, EditorEventManager eventManager) {
        final List<Diagnostic> diagnostics = eventManager.getDiagnostics();
        final Editor editor = eventManager.editor;

        diagnostics.forEach(d -> {
            AnnotationBuilder annotationBuilder = createAnnotationBuilder(editor, holder, d);
            if (annotationBuilder != null) {
                annotationBuilder.create();
            }
        });

        eventManager.setAnonHolder(holder);
    }
```

I'm not entirely sure if we need to retain the `AnnotationBuilder`s in a `List` as we do with the `Annotation`s currently, as I don't entirely understand what the `isCodeActionSyncRequired` case is for or how it is supposed to work. This also seems tricky to deal with in `EventEditorManager.requestAndShowCodeActions()` as it completely relies on having `Annotation`s which contain source positions whereas `AnnotationBuilder`s do not.

Trias:
## Purpose
deprecations can now be rendered. This is a feature of [lsp 3.15](https://microsoft.github.io/language-server-protocol/specification#diagnostic).

also fixes #200

Note that i associated LSP-"Information" with IJ-"Weak warning" and LSP-"hint" with IJ-"information", as "hint" in vscode is rendered invisible like Information is in Intellij. This is just a different choice of semantics.

### DIFF
--- a/src/main/java/org/wso2/lsp4intellij/contributors/annotator/LSPAnnotator.java
+++ b/src/main/java/org/wso2/lsp4intellij/contributors/annotator/LSPAnnotator.java
@@ -15,18 +15,24 @@
  */
 package org.wso2.lsp4intellij.contributors.annotator;
 
+import com.intellij.codeInspection.ProblemHighlightType;
 import com.intellij.lang.annotation.Annotation;
 import com.intellij.lang.annotation.AnnotationHolder;
 import com.intellij.lang.annotation.ExternalAnnotator;
+import com.intellij.lang.annotation.HighlightSeverity;
 import com.intellij.openapi.diagnostic.Logger;
 import com.intellij.openapi.editor.Editor;
 import com.intellij.openapi.util.TextRange;
 import com.intellij.openapi.vfs.VirtualFile;
 import com.intellij.psi.PsiFile;
+import com.intellij.util.SmartList;
 import org.eclipse.lsp4j.Diagnostic;
+import org.eclipse.lsp4j.DiagnosticSeverity;
+import org.eclipse.lsp4j.DiagnosticTag;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 import org.wso2.lsp4intellij.IntellijLanguageClient;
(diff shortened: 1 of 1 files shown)


NipunaRanasinghe (commented):
@Trias Lets resolve conflicts

Trias:
yep.. missed that. conflict is merged.

NipunaRanasinghe (approved):

PR #247, opened 2021-02-24
fixing some minor issues

SUBYEAR:
## Purpose
Some problems are found in using the plugin
1. using the deprecated method
    image
2. there exist some NPE problems 
3. some fields and method declared by `protected` will be easy to expand

### DIFF
--- a/src/main/java/org/wso2/lsp4intellij/client/languageserver/LSPServerStatusWidget.java
+++ b/src/main/java/org/wso2/lsp4intellij/client/languageserver/LSPServerStatusWidget.java
@@ -88,7 +88,7 @@ public class LSPServerStatusWidget implements StatusBarWidget {
             widgetIDs.put(project, list);
         }
 
-        statusBar.addWidget(widget, "before " + widgetIDs.get(project).get(0));
+        statusBar.addWidget(widget, "before " + widgetIDs.get(project).get(0), project);
         widgetIDs.get(project).add(0, widget.ID());
         return widget;
     }
--- a/src/main/java/org/wso2/lsp4intellij/editor/DocumentEventManager.java
+++ b/src/main/java/org/wso2/lsp4intellij/editor/DocumentEventManager.java
@@ -105,6 +105,11 @@ public class DocumentEventManager {
             CharSequence newText = event.getNewFragment();
             int offset = event.getOffset();
             int newTextLength = event.getNewLength();
+            Set<EditorEventManager> managersForUri = EditorEventManagerBase.managersForUri(FileUtils.documentToUri(document));
+            if (managersForUri == null || managersForUri.isEmpty()) {
+                LOG.warn("no manager associated with uri");
+                return;
+            }
             EditorEventManager editorEventManager = EditorEventManagerBase.managersForUri(FileUtils.documentToUri(document)).iterator().next();
             if (editorEventManager == null) {
                 LOG.warn("no editor associated with document");
@@ -112,6 +117,9 @@ public class DocumentEventManager {
             }
             Editor editor = editorEventManager.editor;
             Position lspPosition = DocumentUtils.offsetToLSPPos(editor, offset);
+            if (lspPosition == null) {
+                return;
+            }
             int startLine = lspPosition.getLine();
             int startColumn = lspPosition.getCharacter();
             CharSequence oldText = event.getOldFragment();
--- a/src/main/java/org/wso2/lsp4intellij/editor/EditorEventManager.java
+++ b/src/main/java/org/wso2/lsp4intellij/editor/EditorEventManager.java
@@ -1149,7 +1149,7 @@ public class EditorEventManager {
             return null;
         }
         if (editor.isDisposed()) {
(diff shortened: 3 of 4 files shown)


src/main/java/org/wso2/lsp4intellij/client/languageserver/LSPServerStatusWidget.java
```diff
@@ -88,7 +88,7 @@ public static LSPServerStatusWidget createWidgetFor(LanguageServerWrapper wrappe
             widgetIDs.put(project, list);
         }
 
-        statusBar.addWidget(widget, "before " + widgetIDs.get(project).get(0));
+        statusBar.addWidget(widget, "before " + widgetIDs.get(project).get(0), project);
```
SUBYEAR, line 91: the method without `project` parameter is deprecated

src/main/java/org/wso2/lsp4intellij/editor/DocumentEventManager.java
```diff
@@ -105,13 +105,22 @@ public void documentChanged(DocumentEvent event) {
...
             int offset = event.getOffset();
             int newTextLength = event.getNewLength();
-            EditorEventManager editorEventManager = EditorEventManagerBase.managersForUri(FileUtils.documentToUri(document)).iterator().next();
+            Set<EditorEventManager> managersForUri = EditorEventManagerBase.managersForUri(
+                FileUtils.documentToUri(document));
+            if (managersForUri == null || managersForUri.isEmpty()) {
+                LOG.warn("no manager associated with uri");
+                return;
+            }
+            EditorEventManager editorEventManager = managersForUri.iterator().next();
```
SUBYEAR, line 114: fix potential NPE problem

src/main/java/org/wso2/lsp4intellij/editor/DocumentEventManager.java
```diff
@@ -105,13 +105,22 @@ public void documentChanged(DocumentEvent event) {
...
             if (editorEventManager == null) {
                 LOG.warn("no editor associated with document");
                 return;
             }
             Editor editor = editorEventManager.editor;
             Position lspPosition = DocumentUtils.offsetToLSPPos(editor, offset);
+            if (lspPosition == null) {
+                return;
+            }
```
SUBYEAR, line 123: fix potential NPE problem

src/main/java/org/wso2/lsp4intellij/editor/EditorEventManager.java
```diff
@@ -181,8 +181,8 @@
     public LanguageServerWrapper wrapper;
     private Project project;
     private TextDocumentIdentifier identifier;
-    private EditorMouseListener mouseListener;
-    private EditorMouseMotionListener mouseMotionListener;
+    protected EditorMouseListener mouseListener;
+    protected EditorMouseMotionListener mouseMotionListener;
```
SUBYEAR, line 185: easy to expanding a subclass

Trias: not to say that this change is wrong, but are you actually planning on overwriting or accessing these properties? 

If we let anyone extend our classes (which, sure, is technically already possible) we expose ourselves to the [Fragile Base Class Problem](https://de.wikipedia.org/wiki/Fragile_Base_Class_Problem). So i would rather leave these Properties private.

You can already inject custom listeners via the constructor. Is this not sufficient for your use case?

SUBYEAR: As you said `inject custom listeners via the constructor`,  this cann't be implemented. Because these listener is fixed, cann't be changed.
image
----
Yeah, I want to extend the `EditorEventManager` class, and do some change in method `registerListeners`. So I change the fields accessibility as protected

SUBYEAR: > not to say that this change is wrong, but are you actually planning on overwriting or accessing these properties?
> 
> If we let anyone extend our classes (which, sure, is technically already possible) we expose ourselves to the [Fragile Base Class Problem](https://en.wikipedia.org/wiki/Fragile_base_class). So i would rather leave these Properties private.
> 
> You can already inject custom listeners via the constructor. Is this not sufficient for your use case?

Taking your advice, and keep these fields protected

src/main/java/org/wso2/lsp4intellij/editor/EditorEventManager.java
```diff
@@ -1148,6 +1148,10 @@ public Runnable getEditsRunnable(int version, List<TextEdit> edits, String name,
             LOG.warn("Received edits list is null.");
             return null;
         }
+        if (editor.isDisposed()) {
+            LOG.warn("The edit is disposed.");
```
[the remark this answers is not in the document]
NipunaRanasinghe: @SUBYEAR It seems that the above suggestion is not addressed yet

NipunaRanasinghe (approved):

PR #287, opened 2023-01-27
make getWidget() return an optional and handle this appropriately where this method is used

### LINKED ISSUE #286: ensure lsp status widget is treated as optional
**Description:**
with the changes provided in #251 the status widget has effectively become optional by not including the factory in the plugin.xml file.

For example by excluding the following from the plugin.xml
```
        <!-- for displaying the statusbar icon -->
        <statusBarWidgetFactory implementation="org.wso2.lsp4intellij.statusbar.LSPServerStatusWidgetFactory"
                                id="org.wso2.lsp4intellij.statusbar.LSPServerStatusWidgetFactory"
                                order="first" />
```
It is possible to run the LSP without having the status bar widget displayed to the user.

**Describe your problem(s)**

The only issue with this is there are a few places in LanguageServerWrapper which assume the widget is not null.  So if you do not include the widget the entire lsp fails to startup.

**Describe your solution(s)**

Ensure the getWidget method returns an optional to ensure all current and future updates to the widget are handled in an optional way.

johnamSyn:

### DIFF
--- a/src/main/java/org/wso2/lsp4intellij/client/languageserver/wrapper/LanguageServerWrapper.java
+++ b/src/main/java/org/wso2/lsp4intellij/client/languageserver/wrapper/LanguageServerWrapper.java
@@ -255,7 +255,7 @@ public class LanguageServerWrapper {
     }
 
     public void notifyResult(Timeouts timeouts, boolean success) {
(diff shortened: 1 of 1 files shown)


NipunaRanasinghe (approved):
LGTM

PR #295, opened 2023-02-03
prevent NPE of getOpenClose() tristate boolean by converting null to false

### LINKED ISSUE #294: add extra null check for DefaultRequestManager didOpen method to match didClose
**Description:**
add a extra null check to didOpen method so it matches didClose method preventing a potential NPE due to tri state boolean

johnamSyn:


johnamSyn:
@NipunaRanasinghe this is my last pull request for the time being if you can approve this one and then trigger a patch release that would be great!

NipunaRanasinghe (approved):
LGTM

NipunaRanasinghe:
> @NipunaRanasinghe this is my last pull request for the time being if you can approve this one and then trigger a patch release that would be great!

@johnamSyn Sure we'll proceed with the patch release with this. Really appreciate your effort on this so far 👏

PR #310, opened 2023-05-04
docs: Add example for custom initialization options #307

### LINKED ISSUE #307: [Help] How to custom  initialization  options?
**Description:**

Thank you for this great project and your work. 
I have encountered an issue and would like to ask for help. 
My LSP server requires some initialization options, and as a result, it is throwing an error during connection. Can you please advise me on how to send custom initialization options to my LSP server during the connection process?

dongwa:
## Purpose
Resolves #307

## Goals
Just add some example  for custom initialization options

### DIFF
--- a/README.md
+++ b/README.md
@@ -262,6 +262,9 @@ You can use the following concrete class:
   ProcessBuilder process = new ProcessBuilder("path/to/launcher-script.sh");
   IntellijLanguageClient.addServerDefinition(new MyServerDefinition("xxx", processBuilder));
   ```
+  
+  You can refer to [#311](https://github.com/ballerina-platform/lsp4intellij/pull/311) for more details.
+
 
 > **Note:** All of the above implementations will use server stdin/stdout to communicate.
 
--- a/README.md
+++ b/README.md
@@ -200,6 +200,47 @@ You can use the following concrete class:
     ProcessBuilder process = new ProcessBuilder("java","-jar","path/to/language-server.jar");
     new ProcessBuilderServerDefinition("bsl,os", process);
     ```
+
+- **Custom Initialization Options**
+
+    If your LSP server needs some custom initialization options when connecting, 
+    you can define a class that extends `ProcessBuilderServerDefinition` or `RawCommandServerDefinition`, 
+    and then override the `getInitializationOptions` method. 
+    Here's an example:
+
+    ```java
+    public class MyServerDefinition extends ProcessBuilderServerDefinition {
+        
+        /**
+         * Creates a new instance.
+         *
+         * @param ext     The extension.
+         * @param process The process builder instance to be started.
+         */
+        public MyServerDefinition(String ext, ProcessBuilder process) {
+            super(ext, process);
+        }
+        
+        /**
+         * Returns the custom initialization options for the given URI.
+         *
+         * @param uri The URI.
+         * @return The initialization options.
+         */
+        @Override
+        public Object getInitializationOptions(URI uri) {
+            // Return your custom initialization options here.
+        }
+    }
+    ```
+
+    To use your custom server definition class,
+    you can create a new instance of it and pass it to the `addServerDefinition` method of a `IntellijLanguageClient`. For example:
+
+    ```java
+    ProcessBuilder process = new ProcessBuilder("path/to/launcher-script.sh");
+    IntellijLanguageClient.addServerDefinition(new MyServerDefinition("xxx", processBuilder));
+    ```
 > **Note:** All of the above implementations will use server stdin/stdout to communicate.
 
 ![](resources/images/lang-server-connect.gif)
--- a/README.md
+++ b/README.md
@@ -201,46 +201,68 @@ You can use the following concrete class:
     new ProcessBuilderServerDefinition("bsl,os", process);
     ```
 
-- **Custom Initialization Options**
+- **Custom Initialization Options Or Initialization Params**
 
     If your LSP server needs some custom initialization options when connecting, 
-    you can define a class that extends `ProcessBuilderServerDefinition` or `RawCommandServerDefinition`, 
-    and then override the `getInitializationOptions` method. 
+    you can define a class that extends `ProcessBuilderServerDefinition` or `RawCommandServerDefinition`. 
+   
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -200,6 +200,47 @@ You can use the following concrete class:
...
+        
+        /**
+         * Returns the custom initialization options for the given URI.
+         *
+         * @param uri The URI.
+         * @return The initialization options.
+         */
+        @Override
+        public Object getInitializationOptions(URI uri) {
+            // Return your custom initialization options here.
```
payne911, line 232: I think this should provide an actual example. For example: how would someone set the `InitializeParams#clientInfo` field's value?

payne911 (changes requested):

payne911:
This seems to only provide the ability to assign a value for `initializationOptions`.

I think the library needs to instead provide a way to set the value of any field in this `InitializeParams` class.

dongwa:
> 
Yes, it only handles InitializationOptions, not InitializeParams. You would be better off opening another issue or pull request to address your question.

NipunaRanasinghe:
> > 
> 
> Yes, it only handles InitializationOptions, not InitializeParams. You would be better off opening another issue or pull request to address your question.

I feel like having to create your own process builder just to pass InitializationOptions might not be a good design. Probably we can expose APIs from the `LanguageServerDefinition` to set  both `InitializationOptions` and `InitializeParams`. 

@dongwa @payne911 Any thoughts?

dongwa:
> > > 
> > 
> > 
> > Yes, it only handles InitializationOptions, not InitializeParams. You would be better off opening another issue or pull request to address your question.
> 
> I feel like having to create your own process builder just to pass InitializationOptions might not be a good design. Probably we can expose APIs from the `LanguageServerDefinition` to set both `InitializationOptions` and `InitializeParams`.
> 
> @dongwa @payne911 Any thoughts?

Yes it's better,but now

> 

Yes, if `LanguageServerDefinition` has an API to set `InitializationOptions` and `InitializeParams`, I think it would be better. I am glad to help, but it would not be easy. We need a server to test.

payne911:
# What I've tried
Nothing worked, unfortunately. :(

### CustomServerWrapper
I've been trying to figure out how to provide my own ServerWrapper via the ExtensionManager, but it seems like that's not possible.

### Override `LanguageServer#initialize`
Then instead in my custom RequestManager (which is registered via my ExtensionManager) I tried:
```kotlin
override fun initialize(params: InitializeParams): CompletableFuture<InitializeResult> {
  // do as I wish to modify those params
  return super.initialize(params)
}
```
... but somehow I can't get this to work. (Yet I do have other methods which I've successfully overridden in my Request Manager.)
Shouldn't the above effectively allow us to "post-process" the InitializeParams as we wish before the `initialize` method request is sent to the server? The relevant code pointer:
https://github.com/ballerina-platform/lsp4intellij/blob/d0ca56be2c65d63dfac405d3811f3b67ca372a54/src/main/java/org/wso2/lsp4intellij/client/languageserver/wrapper/LanguageServerWrapper.java#L533

---

# Proposition
@NipunaRanasinghe I believe that providing a method like `modifyInitializeParam(initParams: InitializeParams)` would be good enough:

```
    public void modifyInitializeParam(InitializeParams params) {
        // nothing in here
    }

    public void start() {
        // stuff before
                modifyInitializeParam(initParams); // this gets added
                initializeFuture = languageServer.initialize(initParams).thenApply(res -> {
        // stuff after
    }
```

People could override this new method and modify the pre-processing that this library already does. This way, it's easy to keep part of the automation we already benefit from, while also giving all the possible freedom to the library's users.

payne911:
@NipunaRanasinghe see #311

NipunaRanasinghe (commented):
@dongwa Is it possible to update the PR with changes introduced by https://github.com/ballerina-platform/lsp4intellij/pull/311?

dongwa:
> @dongwa Is it possible to update the PR with changes introduced by #311?

ok, give me some times

NipunaRanasinghe:
> > @dongwa Is it possible to update the PR with changes introduced by #311?
> 
> ok, give me some times

Great! thanks

dongwa:
Okay, I have updated it. Please help me review it.
@NipunaRanasinghe @payne911

NipunaRanasinghe (approved):

README.md
```diff
@@ -200,6 +200,69 @@ You can use the following concrete class:
...
+  }
+  ```
+
+  To use your custom server definition class,
+  you can create a new instance of it and pass it to the `addServerDefinition` method of a `IntellijLanguageClient`. For example:
+
+  ```java
+  ProcessBuilder process = new ProcessBuilder("path/to/launcher-script.sh");
+  IntellijLanguageClient.addServerDefinition(new MyServerDefinition("xxx", processBuilder));
+  ```
```
payne911, line 264:
```suggestion
- **Custom Initialization Params**

If your LSP server needs some custom initialization options when connecting, you can define a class that extends `ProcessBuilderServerDefinition` or `RawCommandServerDefinition`, and then override the `customizeInitializeParams` method to modify any property of the `InitializeParams`.

Here's an example:

```java
public class MyServerDefinition extends ProcessBuilderServerDefinition {
    public MyServerDefinition(String ext, ProcessBuilder process) {
        super(ext, process);
    }

    @Override
    public void customizeInitializeParams(InitializeParams params) {
        params.clientInfo = new ClientInfo("MyName", "MyVersion");
    }
}
```

Finally, assign your class as a ServerDefinition:

```java
ProcessBuilder process = new ProcessBuilder("path/to/launcher-script.sh");
IntellijLanguageClient.addServerDefinition(new MyServerDefinition("xxx", processBuilder));
```

You can refer to #311 for more details.
```

payne911 (changes requested):
GitHub's code suggestion feature does not seem to play well with the markdown code notation. But you'll get the gist of it.

I believe a readme should be light: we should avoid showing lengthy examples. Minimal examples, with reference points is preferable. The JavaDoc provided by the library should be sufficient if users are trying to dig deeper for their understanding.

NipunaRanasinghe (approved):

PR #311, opened 2023-05-10
Allow users to customize the InitializeParams object for the "initialize" method

payne911:
## Purpose
There was no way for users to provide their own values for this object. Now they can by simply overriding this new method.
Feel free to modify this PR directly before merging.

## Goals
Increase flexibility.

## Approach
Usage is as follow:
```kotlin
class MyServerDefinition(ext: String, command: Array<String>) :
    RawCommandServerDefinition(ext, command) {

  override fun customizeInitializeParams(params: InitializeParams) {
    val applicationInfo: ApplicationInfo? =
        ApplicationManager.getApplication().getService(ApplicationInfo::class.java)
    val ideVersion: String? = applicationInfo?.fullVersion
    val ideName = ApplicationNamesInfo.getInstance().fullProductName
    params.clientInfo = ClientInfo(ideName, ideVersion)
  }
}
```

### DIFF
--- a/src/main/java/org/wso2/lsp4intellij/client/languageserver/serverdefinition/LanguageServerDefinition.java
+++ b/src/main/java/org/wso2/lsp4intellij/client/languageserver/serverdefinition/LanguageServerDefinition.java
@@ -77,6 +77,15 @@ public class LanguageServerDefinition {
     public Object getInitializationOptions(URI uri) {
         return null;
     }
+    
+    /**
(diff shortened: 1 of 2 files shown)


NipunaRanasinghe (approved):
This approach looks good! 

IMO we can deprecate the `getInitializationOptions ()` API along with the new one, as the users will now have the full control over the initialize params (so they can directly modify initialization options, if they want to). I'll send those changes in a separate PR, after merging this

PR #228, opened 2020-12-27
fix those lspoutofbounds

swissiety:
- [x] fix outofbounds for linecount -> off by one
- [x] add min bound check
- [x] fix offsettolsp()
- [x] log.info to debug 

closes #198

PR #314, opened 2023-05-26
Sanitize the line and column values if they have been reported as negative when converting the to an LSP position

cpedlar-syn:
## Purpose
LSP locations are defined as uint in the specification and the computation for the column values due to the tabs is not correct.  Passing a negative value to an LSP server can cause them to report errors.

## Goals
Prevent passing negative values to LSP servers

## Approach
This particular change prevents the passing of the value, but it may be better to not pass a position at all to the LSP servers instead of a sanitized one.  Additionally, the real fix is probably resolving the computation for the column value but at this time I'm not confident what that may be.  I'll try to take another look at the computation for the column and see if I can determine why it is incorrect in at least some cases.

## Samples
I have a method with the following format in some test code.  Hovering over the word "bad" in the method will cause the incorrect column to be reported.
```
public class Foo {
	public static String giveNull() {
		String bad = null;
		bad += "foo";   // bad
		return bad;
	}
}
```

Here is a screenshot of the data at that time.
image

### DIFF
--- a/src/main/java/org/wso2/lsp4intellij/utils/DocumentUtils.java
+++ b/src/main/java/org/wso2/lsp4intellij/utils/DocumentUtils.java
@@ -117,8 +117,30 @@ public class DocumentUtils {
             int tabs = StringUtil.countChars(lineTextBeforeOffset, '\t');
             int tabSize = getTabSize(editor);
             int column = lineTextBeforeOffset.length() - tabs * (tabSize - 1);
+
+            // Language server positions must be greater than or equal to 0 according to the LSP specification
+            // https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#position
+            // It is possible for the column (and perhaps line?) to be negative values here
+            // If the value of either is negative, sanitize it to zero
+            line = sanitizeLspPosition(line);
+            column = sanitizeLspPosition(column);
+
             return new Position(line, column);
         });
+    }
+
+    /**
+     * Returns the value if the value is greater than or equal to 0.
+     * If the value is negative returns 0.
+     * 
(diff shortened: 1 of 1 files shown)


nixel2007:
I'm not sure that incrementing column number by tab * tabSize is a correct behavior at all. Lsp position is a position in _characters_. Tab is a character.

nixel2007:
Please review

NipunaRanasinghe (commented):
@cpedlar-syn as our client library itself derive the column number using the offset, sanitising the errornous (negative) outputs might hide the real issue and ends up having unexpected results.

I also feel like we can completely get rid of tabsize based calculations in here, which seems to be a regression introduced with https://github.com/ballerina-platform/lsp4intellij/pull/228.

NipunaRanasinghe:
@nixel2007 Appreciate your review on https://github.com/ballerina-platform/lsp4intellij/pull/323. 

@cpedlar-syn will you be able to verify whether your issue also resolves with the changes proposed with the above PR?

cpedlar-syn:
@NipunaRanasinghe Yes, I can try to adjust the change to just remove the tab based calculations.  I also thought that may be the more correct solution.

I'll try to find some time today to make the changes and circle back, however, if I don't get to it today I won't have another opportunity for ~2 weeks but I will try to make the change as soon as possible.

nixel2007:
Ping

NipunaRanasinghe:
> Ping

@nixel2007 we can simply close this if https://github.com/ballerina-platform/lsp4intellij/pull/323 looks good.

cpedlar-syn:
Closing this PR since the change was merged via another PR.

PR #317, opened 2023-06-04
Fix NPE when handling Hover Request

jczuchnowski:
## Purpose
In Hover Request the server can respond with either a boolean or HoverOptions. In the current implementation in case when the server responds with HoverOptions the Either.left is `null` and this causes a NullPointerException.

## Goals
The goal is to correctly handle response from the server.

## Approach
A simple reverse of the order of `||` check will prevent the NPE from happening.

## User stories
N/A

## Release note
Fix NPE on Hover Request when the response contains HoverOptions.

## Documentation
N/A

## Training
N/A

## Certification
N/A

## Marketing
N/A

## Automation tests
 N/A

## Security checks
 - Followed secure coding standards in http://wso2.com/technical-reports/wso2-secure-engineering-guidelines? yes
 - Ran FindSecurityBugs plugin and verified report? no
 - Confirmed that this PR doesn't commit any keys, passwords, tokens, usernames, or other secrets? yes

## Samples
N/A

## Related PRs
N/A

## Migrations (if applicable)
N/A

## Test environment
N/A
 
## Learning
N/A

### DIFF
--- a/src/main/java/org/wso2/lsp4intellij/client/languageserver/requestmanager/DefaultRequestManager.java
+++ b/src/main/java/org/wso2/lsp4intellij/client/languageserver/requestmanager/DefaultRequestManager.java
@@ -427,7 +427,7 @@ public class DefaultRequestManager implements RequestManager {
             try {
                 return
                         Optional.ofNullable(serverCapabilities.getHoverProvider())
-                                .map(e -> e.getLeft() || e.getRight() != null).orElse(false) ?
+                                .map(e -> e.getRight() != null || e.getLeft()).orElse(false) ?
                                 textDocumentService.hover(params) : null;
 
             } catch (Exception e) {
--- a/src/main/java/org/wso2/lsp4intellij/client/languageserver/requestmanager/DefaultRequestManager.java
+++ b/src/main/java/org/wso2/lsp4intellij/client/languageserver/requestmanager/DefaultRequestManager.java
@@ -427,7 +427,7 @@ public class DefaultRequestManager implements RequestManager {
             try {
                 return
                         Optional.ofNullable(serverCapabilities.getHoverProvider())
-                                .map(e -> e.getRight() != null || e.getLeft()).orElse(false) ?
+                                .map(e -> e.getRight() != null || (e.getLeft() != null && e.getLeft())).orElse(false) ?
                                 textDocumentService.hover(params) : null;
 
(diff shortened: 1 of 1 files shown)


src/main/java/org/wso2/lsp4intellij/client/languageserver/requestmanager/DefaultRequestManager.java
```diff
@@ -427,7 +427,7 @@ public CompletableFuture<Hover> hover(HoverParams params) {
             try {
                 return
                         Optional.ofNullable(serverCapabilities.getHoverProvider())
-                                .map(e -> e.getLeft() || e.getRight() != null).orElse(false) ?
+                                .map(e -> e.getRight() != null || e.getLeft()).orElse(false) ?
```
NipunaRanasinghe, line 430:
```suggestion
                                .map(e -> e.getRight() != null || (e.getLeft() != null && e.getLeft())).orElse(false) ?
```

NipunaRanasinghe: @jczuchnowski since `e.getLeft()` is a a `Boolean` object, we still might have to consider the possibility of it being null IMO. So shall we do the checks in the above way?

NipunaRanasinghe (changes requested):

src/main/java/org/wso2/lsp4intellij/client/languageserver/requestmanager/DefaultRequestManager.java
jczuchnowski: If both left and right is null then that would be a violation of the protocol. So I suppose there are two solutions to consider:
1. We assume false (as per your suggestion)
2. We throw a meaningful exception about the broken protocol.

I'm just wondering if 1. isn't misleading somehow. Eg. developer makes a mistake - wants to return true, but returns null. There's no error, everything works, but not as it's supposed to. Not sure if that's a valid concern, though. I you think defaulting to false is an acceptable solution, then I'm for it.

jczuchnowski: Ok, let's go with false by default.

NipunaRanasinghe (approved):

PR #332, opened 2023-07-31
Ignore URIs that do not start with 'file:'

akashnagesh:
## Purpose
Ignore URIs that do not start with `file://`

Using URIs that have other schemes could lead to `IllegalArgumentException`

**Sample stack-trace:** 

> java.lang.IllegalArgumentException: URI scheme is not "file"
at java.base/java.io.File.<init>(File.java:423)
at org.wso2.lsp4intellij.utils.FileUtils.virtualFileFromURI(FileUtils.java:153)
at org.wso2.lsp4intellij.utils.FileUtils.getAllOpenedEditorsForUri(FileUtils.java:87)
at org.wso2.lsp4intellij.client.languageserver.wrapper.LanguageServerWrapper.connect(LanguageServerWrapper.java:748)

The above exception happens when opening a de-complied source file that have scheme `JRT://`

### DIFF
--- a/src/main/java/org/wso2/lsp4intellij/utils/FileUtils.java
+++ b/src/main/java/org/wso2/lsp4intellij/utils/FileUtils.java
@@ -150,11 +150,15 @@ public class FileUtils {
 
     public static VirtualFile virtualFileFromURI(String uri) {
         try {
-            return LocalFileSystem.getInstance().findFileByIoFile(new File(new URI(sanitizeURI(uri))));
+            URI sanitizedUri = new URI(sanitizeURI(uri));
+            if (sanitizedUri.getScheme().startsWith(URI_FILE_BEGIN)) {
+                return LocalFileSystem.getInstance().findFileByIoFile(new File(sanitizedUri));
+            }
+            LOG.debug(String.format("URI %s is not a file", uri));
         } catch (URISyntaxException e) {
             LOG.warn(e);
-            return null;
         }
+        return null;
     }
 
     /**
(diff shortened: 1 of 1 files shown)


src/main/java/org/wso2/lsp4intellij/utils/FileUtils.java
```diff
@@ -249,12 +253,7 @@ public static String sanitizeURI(String uri) {
      * @return The virtual file
      */
     public static VirtualFile URIToVFS(String uri) {
-        try {
-            return LocalFileSystem.getInstance().findFileByIoFile(new File(new URI(sanitizeURI(uri))));
-        } catch (URISyntaxException e) {
-            LOG.warn(e);
-            return null;
-        }
+        return virtualFileFromURI(uri);
```
akashnagesh, line 256: This method appears to be a duplicate of `virtualFileFromURI()` unless I am missing something.

nixel2007:
I completely disagree with the purpose of this pr. uri and lsp are schema agnostic, and we must not limit implementation in any way

akashnagesh:
> uri and lsp are schema agnostic, and we must not limit implementation in any way

I agree with this, though we seem to specifically check for `file://` while sanitizing URI: https://github.com/ballerina-platform/lsp4intellij/blob/3fedd435ae526c568a46f68816fa1a595fd3a921/src/main/java/org/wso2/lsp4intellij/utils/FileUtils.java#L214C37-L214C37

Any suggestions on how we can fix this @nixel2007 ?

Issue #347, opened 2024-03-28
InvocationTargetException on the startup of the client


ThemiraChathumina:
**Description:**

Getting error java.lang.reflect.InvocationTargetException when starting the client for Ballerina 2201.8.4 (Swan Lake Update 8) Language Server.

```
java.lang.RuntimeException: java.lang.reflect.InvocationTargetException
	at org.eclipse.lsp4j.jsonrpc.services.GenericEndpoint.lambda$null$0(GenericEndpoint.java:67)
	at org.eclipse.lsp4j.jsonrpc.services.GenericEndpoint.request(GenericEndpoint.java:120)
	at org.eclipse.lsp4j.jsonrpc.RemoteEndpoint.handleRequest(RemoteEndpoint.java:261)
	at org.eclipse.lsp4j.jsonrpc.RemoteEndpoint.consume(RemoteEndpoint.java:190)
	at org.wso2.lsp4intellij.client.languageserver.wrapper.MessageHandler.lambda$apply$0(MessageHandler.java:45)
	at org.eclipse.lsp4j.jsonrpc.json.StreamMessageProducer.handleMessage(StreamMessageProducer.java:194)
	at org.eclipse.lsp4j.jsonrpc.json.StreamMessageProducer.listen(StreamMessageProducer.java:94)
	at org.eclipse.lsp4j.jsonrpc.json.ConcurrentMessageProcessor.run(ConcurrentMessageProcessor.java:113)
	at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:539)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1136)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635)
	at java.base/java.lang.Thread.run(Thread.java:833)
Caused by: java.lang.reflect.InvocationTargetException
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
[...]

PR #348, opened 2024-03-28
Add IDE notifications for LSP progress update messages

ThemiraChathumina:
## Purpose
Adds editor event log tab notifications for LSP messages that display progress update information. Display notifications for start and end of a work done progress. This also fixes the issue https://github.com/ballerina-platform/lsp4intellij/issues/347 that happens due to not supporting LSP progress indication messages.

## Approach
```createProgress(WorkDoneProgressCreateParams params)```: This method is triggered upon receiving a _**window/workDoneProgress/create**_ request from the language server. It extracts the progress token from the request parameters, which can be either a String or a numeric value. This token uniquely identifies the progress report session. The method then initializes a progress notification item with a default title and an empty message, associating it with the token in a map for later reference. This setup prepares the IDE to handle upcoming progress notifications associated with this token.

```notifyProgress(ProgressParams params)```: Activated by the _**$/progress**_ notification from the language server, this method manages the display of progress notifications within the IDE. It first retrieves the progress token from the notification parameters, similar to createProgress. Using this token, it fetches the corresponding progress notification item from the map. The method then updates the title and message based on the type of progress notification received (e.g., WorkDoneProgressBegin, WorkDoneProgressEnd). 

Notification will be associated with the extension name provided when initializing the language client.

## Samples

image

image

### DIFF
--- a/src/main/java/org/wso2/lsp4intellij/client/DefaultLanguageClient.java
+++ b/src/main/java/org/wso2/lsp4intellij/client/DefaultLanguageClient.java
@@ -334,7 +334,7 @@ public class DefaultLanguageClient implements LanguageClient {
             }
         }
         if (extension != null) {
-            title = " (" + extension + " extension" + ") " + title;
+            title = " [" + extension + " extension" + "] " + title;
         }
         Notification notification =
                 notificationGroup.createNotification(title, message, NotificationType.INFORMATION, null);
--- a/src/main/java/org/wso2/lsp4intellij/client/DefaultLanguageClient.java
+++ b/src/main/java/org/wso2/lsp4intellij/client/DefaultLanguageClient.java
@@ -293,7 +293,6 @@ public class DefaultLanguageClient implements LanguageClient {
 
     @Override
     public CompletableFuture<Void> createProgress(WorkDoneProgressCreateParams params) {
-        // give a unique identifier if the token is null
         String token;
         if (params.getToken().getLeft() != null) {
             token = params.getToken().getLeft();
--- a/src/main/java/org/wso2/lsp4intellij/client/DefaultLanguageClient.java
+++ b/src/main/java/org/wso2/lsp4intellij/client/DefaultLanguageClient.java
@@ -47,6 +47,7 @@ import org.eclipse.lsp4j.WorkDoneProgressNotification;
 import org.eclipse.lsp4j.WorkspaceFolder;
 import org.eclipse.lsp4j.services.LanguageClient;
 import org.jetbrains.annotations.NotNull;
+import org.wso2.lsp4intellij.client.languageserver.wrapper.LanguageServerWrapper;
 import org.wso2.lsp4intellij.editor.EditorEventManagerBase;
 import org.wso2.lsp4intellij.requests.WorkspaceEditHandler;
 import org.wso2.lsp4intellij.utils.ApplicationUtils;
@@ -335,7 +336,14 @@ public class DefaultLanguageClient implements LanguageClient {
                 title = progressNotificationItems.get(token).getFirst();
             }
         }
-        String extension = ((ServerWrapperBaseClientContext) context).getWrapper().serverDefinition.ext;
+        String extension = null;
+        if (context instanceof ServerWrapperBaseClientContext) {
+            ServerWrapperBaseClientContext serverContext = (ServerWrapperBaseClientContext) context;
+            LanguageServerWrapper wrapper = serverContext.getWrapper();
+            if (wrapper != null && wrapper.serverDefinition != null) {
+                extension = wrapper.serverDefinition.ext;
+            }
+        }
         if (extension != null) {
             title = " [" + extension + " extension" + "] " + title;
         }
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: src/main/java/org/wso2/lsp4intellij/client/DefaultLanguageClient.java [hunk 1]
```diff
@@ -20,27 +20,49 @@
...
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.FutureTask;
 
+import javax.swing.*;
```
NipunaRanasinghe, line 64: Can we avoid the wildcard import in here?
```suggestion
import javax.swing.*;
```

src/main/java/org/wso2/lsp4intellij/client/DefaultLanguageClient.java
```diff
@@ -265,4 +290,54 @@ public void logMessage(MessageParams messageParams) {
...
     }
+
+    @Override
+    public CompletableFuture<Void> createProgress(WorkDoneProgressCreateParams params) {
+        String token;
+        if (params.getToken().getLeft() != null) {
+            token = params.getToken().getLeft();
+        } else {
+            token = params.getToken().getRight().toString();
+        }
```
NipunaRanasinghe, line 301: Shall we use a `else if` here instead, to avoid potential NPE where both `getLeft()` and `getRight()` return null?
```suggestion
        } else {
            token = params.getToken().getRight().toString();
        }
```

src/main/java/org/wso2/lsp4intellij/client/DefaultLanguageClient.java
```diff
@@ -265,4 +290,54 @@ public void logMessage(MessageParams messageParams) {
...
+    public void notifyProgress(ProgressParams params) {
+        NotificationGroup notificationGroup =
+                NotificationGroupManager.getInstance().getNotificationGroup("LSPProgressNotification");
+
+        String token;
+        if (params.getToken().getLeft() != null) {
+            token = params.getToken().getLeft();
+        } else {
+            token = params.getToken().getRight().toString();
+        }
```
NipunaRanasinghe, line 317: ditto

src/main/java/org/wso2/lsp4intellij/client/DefaultLanguageClient.java
```diff
@@ -265,4 +290,54 @@ public void logMessage(MessageParams messageParams) {
...
+            }
+        } else if (progressNotification instanceof WorkDoneProgressEnd) {
+            message = ((WorkDoneProgressEnd) progressNotification).getMessage();
+            if (progressNotificationItems.containsKey(token)) {
+                title = progressNotificationItems.get(token).getFirst();
+            }
+        }
+        if (extension != null) {
+            title = " (" + extension + " extension" + ") " + title;
+        }
```
NipunaRanasinghe, line 338:
```suggestion
        if (extension != null) {
            title = " [" + extension + " extension" + "] " + title;
        }
```

src/main/java/org/wso2/lsp4intellij/client/DefaultLanguageClient.java
```diff
@@ -265,4 +290,59 @@ public void logMessage(MessageParams messageParams) {
     protected final ClientContext getContext() {
         return context;
     }
+
+    @Override
+    public CompletableFuture<Void> createProgress(WorkDoneProgressCreateParams params) {
+        // give a unique identifier if the token is null
```
NipunaRanasinghe, line 296:
```suggestion
```

src/main/java/org/wso2/lsp4intellij/client/DefaultLanguageClient.java
```diff
@@ -52,9 +74,12 @@ public class DefaultLanguageClient implements LanguageClient {
     @NotNull
     private final ClientContext context;
     protected boolean isModal = false;
+    private final String extension;
+    private final HashMap<String, Tuple2<String, String>> progressNotificationItems = new HashMap<>();
 
-    public DefaultLanguageClient(@NotNull ClientContext context) {
+    public DefaultLanguageClient(@NotNull ClientContext context, String extension) {
```
NipunaRanasinghe, line 80: We don't need to pass the extension as a separate contructor param cuz we can get the same via the `context`.

src/main/java/org/wso2/lsp4intellij/client/DefaultLanguageClient.java
```diff
@@ -338,6 +335,7 @@ public void notifyProgress(ProgressParams params) {
                 title = progressNotificationItems.get(token).getFirst();
             }
         }
+        String extension = ((ServerWrapperBaseClientContext) context).getWrapper().serverDefinition.ext;
```
NipunaRanasinghe, line 338: We might need to do an `instanceof` check before directly casting to`ServerWrapperBaseClientContext`.

NipunaRanasinghe (approved):

Issue #278, opened 2022-11-03
IllegalArgumentException error after upgrading to version 0.95.0


maldz3:
**Description:**

Trying to do a simple code completion that works with version 0.94.2:

```
let foo = { val1: "String", val2: 42 }
foo.val2
```

**Affected Versions:**

0.95.0

**Error**

```
java.lang.IllegalArgumentException: Argument for @NotNull parameter 'currentElement' of com/intellij/codeInsight/daemon/impl/B.<init> must not be null
	at com.intellij.codeInsight.daemon.impl.B.$$$reportNull$$$0(B.java)
	at com.intellij.codeInsight.daemon.impl.B.<init>(B.java)
	at com.intellij.codeInsight.daemon.impl.AnnotationHolderImpl.newAnnotation(AnnotationHolderImpl.java:238)
	at org.wso2.lsp4intellij.editor.EditorEventManager.lambda$requestAndShowCodeActions$48(EditorEventManager.java:1504)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1541)
	at org.wso2.lsp4intellij.editor.EditorEventManager.lambda$requestAndShowCodeActions$49(EditorEventManager.java:1466)
	at com.intellij.openapi.application.TransactionGuardImpl.runWithWritingAllowed(TransactionGuardImpl.java:214)
	at com.intellij.openapi.application.TransactionGuardImpl.access$200(TransactionGuardImpl.java:21)
	at com.intellij.openapi.application.TransactionGuardImpl$2.run(TransactionGuardImpl.java:196)
	at com.intellij.openapi.application.impl.ApplicationImpl.runIntendedWriteActionOnCurrentThread(ApplicationImpl.java:805)
	at com.intellij.openapi.application.impl.ApplicationImpl.lambda$invokeLater$4(ApplicationImpl.java:348)
	at com.intellij.openapi.application.impl.FlushQueue.doRun(FlushQueue.java:82)
	at com.intellij.openapi.application.impl.FlushQueue.runNextEvent(FlushQueue.java:131)
	at com.intellij.openapi.application.impl.FlushQueue.flushNow(FlushQueue.java:47)
	at com.intellij.openapi.application.impl.FlushQueue$FlushNow.run(FlushQueue.java:187)
	at java.desktop/java.awt.event.InvocationEvent.dispatch(InvocationEvent.java:313)
	at java.desktop/java.awt.EventQueue.dispatchEventImpl(EventQueue.java:776)
[...]

PR #349, opened 2024-03-28
Suppress the IllegalArgumentException thrown when creating a new annotation

ThemiraChathumina:
## Purpose
Triggering code actions without a diagnostic context currently results in an ```IllegalArgumentException```. This fix will ensure that the corresponding IDE error message is suppressed.

Related to https://github.com/ballerina-platform/lsp4intellij/issues/278

## Approach
Gracefully handle ```IllegalArgumentException``` within a try-catch block. This error will be thrown when creating a new annotation object to display the relevant code action.


NipunaRanasinghe (approved):
Please note that we've temporarily suppressed the exception to omit plugin crash logs, until a proper fix arrieves

Issue #351, opened 2024-04-09
Incorrect code completions on the start of the document


ThemiraChathumina:
**Description:**

When typing at the beginning of a document (offset 0), code completions fail to incorporate the initial characters typed. As a result, when a completion suggestion is accepted, it incorrectly appends to the text initially entered.

![20240409090208online-video-cutter com-ezgif com-video-to-gif-converter](https://github.com/ballerina-platform/lsp4intellij/assets/110608712/f187a152-a5a2-4926-bda2-8eaa2e10465c)

Also getting the following IDE error when inserting a suggestion that is not a code snippet.
```
java.lang.Throwable: Incorrect offsets: startOffset=9, endOffset=16, text length=9
	at com.intellij.openapi.diagnostic.Logger.error(Logger.java:161)
	at com.intellij.openapi.editor.impl.DocumentImpl.createRangeMarker(DocumentImpl.java:507)
	at com.intellij.openapi.editor.Document.createRangeMarker(Document.java:229)
	at com.intellij.codeInsight.template.impl.TemplateState.start(TemplateState.java:321)
	at com.intellij.codeInsight.template.impl.TemplateState.start(TemplateState.java:297)
	at com.intellij.codeInsight.template.impl.TemplateManagerImpl.lambda$startTemplate$2(TemplateManagerImpl.java:171)
	at com.intellij.openapi.command.impl.CoreCommandProcessor.executeCommand(CoreCommandProcessor.java:207)
	at com.intellij.openapi.command.impl.CoreCommandProcessor.executeCommand(CoreCommandProcessor.java:172)
	at com.intellij.openapi.command.impl.CoreCommandProcessor.executeCommand(CoreCommandProcessor.java:162)
	at com.intellij.openapi.command.impl.CoreCommandProcessor.executeCommand(CoreCommandProcessor.java:148)
	at com.intellij.codeInsight.template.impl.TemplateManagerImpl.startTemplate(TemplateManagerImpl.java:175)
	at com.intellij.codeInsight.template.impl.TemplateManagerImpl.startTemplate(TemplateManagerImpl.java:193)
	at com.intellij.codeInsight.template.impl.TemplateManagerImpl.startTemplate(TemplateManagerImpl.java:132)
	at org.wso2.lsp4intellij.editor.EditorEventManager.prepareAndRunSnippet(EditorEventManager.java:1018)
[...]

PR #352, opened 2024-04-09
Fix code completion issue on the document start offset

ThemiraChathumina:
## Purpose
When typing at the start of a document (offset 0), the code completion feature does not capture the initial characters typed. Consequently, accepting a completion suggestion results in it being incorrectly appended to the initially typed text. This fix aims to ensure that suggestions are accurately based on the entire text entered and are inserted correctly. Related to https://github.com/ballerina-platform/lsp4intellij/issues/351

## Approach
**Change the function** ```public String getCompletionPrefix(Editor editor, int offset)```
In the original implementation, the function scans backwards from the offset until it encounters a delimiter. If no delimiter is found, it returns an empty string. However, at the start of the document where there are no preceding delimiters, this approach results in an empty string being returned, failing to capture the text initially typed. To address this, the function should be adjusted to return the captured characters up to the point where a delimiter is encountered, or, if at the start of the document where no delimiters are present, return the text from the start to the current offset.

**Change the function** ```public void prepareAndRunSnippet(String insertText)```
An IDE error throws when trying to insert a suggestion that is not a code snippet from this function. Therefore, if the insertText does not align with recognized code snippet patterns, the function should return immediately, avoiding further processing.

### DIFF
--- a/src/main/java/org/wso2/lsp4intellij/editor/EditorEventManager.java
+++ b/src/main/java/org/wso2/lsp4intellij/editor/EditorEventManager.java
@@ -959,20 +959,16 @@ public class EditorEventManager {
 
     @NotNull
     public String getCompletionPrefix(Editor editor, int offset) {
-        List<String> delimiters = new ArrayList<>(this.completionTriggers);
-        // add whitespace as delimiter, otherwise forced completion does not work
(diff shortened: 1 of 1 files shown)


NipunaRanasinghe (approved):
LGTM

Issue #364, opened 2024-05-10
Version 0.96.0 causes to slow down the IDE.


ThemiraChathumina:
**Description:**
After upgrading to the latest version with the Ballerina language server, none of the features are working and the IDE is lagging. All requests are timing out. This issue is occurring in both IntelliJ 2023 and 2024 versions.

**Steps to reproduce:**

**Affected Versions:**

**OS, DB, other environment details and versions:**

**Related Issues (optional):**

**Suggested Labels (optional):**

**Suggested Assignees (optional):**

NipunaRanasinghe:
Closing with https://github.com/ballerina-platform/lsp4intellij/pull/365

PR #365, opened 2024-05-10
Fix lagging issues in the version 0.96.0

ThemiraChathumina:
## Purpose
This PR resolves the lagging issues and IDE errors that occur when using the library with IntelliJ versions 2023 and 2024. Related to https://github.com/ballerina-platform/lsp4intellij/issues/364

## Approach
The root cause of the lag is the checking of each message passed between the server and client. After removing the relevant try-catch block, a null pointer exception occurs for progress notification messages. This issue arises because the return value of the method `public CompletableFuture<Void> createProgress(WorkDoneProgressCreateParams params)` is `null`. After adding a proper type-matching return value, the null pointer error and lagging issue no longer occur.


NipunaRanasinghe (approved):
LGTM

Issue #371, opened 2024-05-20
Some code snippets are entered incorrectly.


ThemiraChathumina:
**Description:**
Below snippets are entered incorrectly in Ballerina language server.

![image](https://github.com/ballerina-platform/lsp4intellij/assets/110608712/a938d0d3-e7b6-43cb-b168-f9889be7ac52)

extra `$` marks and incorrect placeholders are added in the snippet text.

![image](https://github.com/ballerina-platform/lsp4intellij/assets/110608712/9d3f63dc-d995-4557-b286-a58d0a0bf6be)

**Steps to reproduce:**

**Affected Versions:**

**OS, DB, other environment details and versions:**

**Related Issues (optional):**

**Suggested Labels (optional):**

**Suggested Assignees (optional):**

PR #373, opened 2024-05-20
Fix incorrect code snippets issue

ThemiraChathumina:
## Purpose
Fixes https://github.com/ballerina-platform/lsp4intellij/issues/371

## Approach
Updated the regex pattern to match nested braces as well.


NipunaRanasinghe (approved):
LGTM

